PR #5, opened 2022-07-21
feat/add delegated api support

colek42:
This adds support for SPIRE delegated API.  This allows withness to request certs on behalf of the workload it is running.


signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
 	return cryptoutil.NewSigner(svid.PrivateKey, cryptoutil.SignWithIntermediates(svid.Certificates[1:]), cryptoutil.SignWithCertificate(svid.Certificates[0]))
 }
+
+func DelgatedSigner(ctx context.Context, socketPath string, workloadPath string) (cryptoutil.Signer, error) {
+	client, err := getDelgatedClient(ctx, socketPath)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("#####: %s", socketPath)
```
mikhailswift, line 64: remove fmt.Prints

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+	binPath, err := exec.LookPath(workloadPath)
+	if err != nil {
+		return nil, err
+	}
+
+	hasher := sha256.New()
+	s, err := ioutil.ReadFile(binPath)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		os.Exit(1)
```
mikhailswift, line 75: remove fmt.Print and don't call os.Exit from library code

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+
+	fmt.Printf("#####: %s", socketPath)
+
+	binPath, err := exec.LookPath(workloadPath)
+	if err != nil {
+		return nil, err
+	}
+
+	hasher := sha256.New()
+	s, err := ioutil.ReadFile(binPath)
```
mikhailswift, line 72: prefer os.ReadFile, ioutil is deprecated

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		os.Exit(1)
+	}
+
+	_, err = hasher.Write(s)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		os.Exit(1)
```
mikhailswift, line 81: same as above, remove print and don't call os exit from library code

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+	resp, err := stream.Recv()
+	if err != nil {
+		return nil, err
+	}
+
+	svids := resp.X509Svids
+
+	svid := svids[0]
+
+	fmt.Printf("### %s", svid.X509Svid)
```
mikhailswift, line 113: remove

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+
+	var intermediates []*x509.Certificate
+
+	for _, cert := range chain {
+		cert, err := x509.ParseCertificate(cert)
+		if err != nil {
+			return nil, err
+		}
+
+		fmt.Printf("\n%s", cert.URIs[0])
```
mikhailswift, line 132: remove

mikhailswift (changes requested):
is there anyway to test this functionality?

mikhailswift:
commit structure should be `feat: xxxx` for conventional commits and changelog generator to work appropriately

colek42:
> is there anyway to test this functionality?

The SPIRE agent needs to be set up with delegated API to test it.  I have it working in scribe, but we should probably find a way to test SPIRE stuff in our open source.

signer/spiffe/spiffe.go
mikhailswift: In actuality we should probably be using os.Open and io.Copy instead of reading the entire file into memory.

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := stream.Recv()
+	if err != nil {
+		return nil, err
+	}
+
+	svids := resp.X509Svids
```
mikhailswift, line 109: svids isn't used outside of the line below this, so this can just be `svid := resp.X509Svids[0]`.

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+
+	svid := svids[0]
+
+	fmt.Printf("### %s", svid.X509Svid)
+
+	chain := svid.X509Svid.CertChain
```
mikhailswift, line 115: similarly this is only used in the loop below, so can be `for _, cert := range svid.X509Svid.CertChain`

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
...
+
+	privateBytes := svid.X509SvidKey
```
mikhailswift, line 117: And same here

signer/spiffe/spiffe.go
```diff
@@ -45,3 +54,96 @@ func Signer(ctx context.Context, socketPath string) (cryptoutil.Signer, error) {
 
 	return cryptoutil.NewSigner(svid.PrivateKey, cryptoutil.SignWithIntermediates(svid.Certificates[1:]), cryptoutil.SignWithCertificate(svid.Certificates[0]))
 }
+
+func DelgatedSigner(ctx context.Context, socketPath string, workloadPath string) (cryptoutil.Signer, error) {
```
mikhailswift, line 58: we probably should be checking array lengths before accessing them in this function to prevent panics

colek42:
I think this would make more sense as a "blocklist" of hashes we do not trust.

mikhailswift:
Not sure what you mean on the blocklist comment?

What's the status of this PR?

PR #6, opened 2022-08-01
feat: add dsse verifier threshold

mikhailswift:
Adds the ability to require at least a specified number of verifiers to
successfully verify a DSSE envelope.

Also fixes witness issue #194 where multiple public keys within a policy
will cause all verification to fail. This was due to the DSSE verify
function returning as soon as a single verifier fails. Now the code
appropriately tries to verify against all passed in verifiers and only
throws an error if no verifier successfully verifies the envelope or the
threshold is unmet.

Signed-off-by: Mikhail Swift <mikhail@testifysec.com>

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,3 +1,17 @@
+# Copyright 2022 The Witness Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


mikhailswift:
Fixes https://github.com/testifysec/witness/issues/194

colek42 (approved):
LGTM!

PR #8, opened 2022-08-23
feature/tetragon bpfsupport

colek42:
This is a draft PR that adds Tetragon BPF support.

This adds two flags

1. `watch-prefix` sets a trace on all files
2.  `tetragon-address` this should be set to the tetragon agent

- We need to add support for containers.  This can be done by getting the current namespace and setting a trace on that if we are inside a container.

- I changed some of the data structures for the tracing.  Once we agree on the data format, we need to modify the ptrace functionality to match it.  This probably needs to happen over a few of PRs.

1. Change data structure
2. Modify ptrace code
4. Add BPF


attestation/commandrun/commandrun.go
```diff
@@ -154,21 +225,46 @@ func (r *CommandRun) runCmd(ctx *attestation.AttestationContext) error {
...
 
+	runtime.LockOSThread()
+
 	if err := c.Start(); err != nil {
 		return err
 	}
 
-	var err error
+	if r.tetragonAddress != "" {
+		syscall.Kill(c.Process.Pid, syscall.SIGSTOP)
```
mikhailswift, line 235: This seems racy -- the process may already well be off to the races by time it gets the SIGSTOP.  We likely want to do something similar to how we start the process with trace like [here](https://github.com/testifysec/go-witness/blob/06e9a338aa4a6f45de1c03c0b3adba0a1d271aad/attestation/commandrun/tracing_linux.go#L49) where the child process will call PTRACE_TRACEME before execve'ing to the new process.

We may not want to use TRACEME exactly since to get the process to go we'd have to attach and then detach a tracer to get the process to go after tetragon is up and running -- but we're going to need to modify all of this to make sure we're starting the trace after the new process is created but before execve is called.

attestation/commandrun/commandrun.go
```diff
@@ -154,21 +225,46 @@ func (r *CommandRun) runCmd(ctx *attestation.AttestationContext) error {
...
+		tc, err := NewTC(ctx, r, c.Process.Pid)
+		if err != nil {
+			return err
+		}
+		err = tc.Start()
+		if err != nil {
+			return err
+		}
+		defer tc.Stop(r)
+		time.Sleep(time.Second * 1)
```
mikhailswift, line 245: Similarly we will need to iron out any timing issues to make sure this is all dependable and not reliant on inconsistent timings.

attestation/commandrun/commandrun.go
```diff
@@ -81,32 +91,89 @@ func New(opts ...Option) *CommandRun {
...
-	Stderr    string        `json:"stderr,omitempty"`
-	ExitCode  int           `json:"exitcode"`
-	Processes []ProcessInfo `json:"processes,omitempty"`
-
+	Cmd                  []string      `json:"cmd"`
+	Stdout               string        `json:"stdout,omitempty"`
+	Stderr               string        `json:"stderr,omitempty"`
+	ExitCode             int           `json:"exitcode"`
+	Processes            []ProcessInfo `json:"processes,omitempty"`
+	Files                []FileInfo    `json:"files,omitempty"`
```
mikhailswift, line 100: we'll probably want this to be a `map[string]FileInfo` so we can look into it by path easily

mikhailswift:
The struct changes look sane for the most part, I'll go over them in more detail later.

I am unsure if we are going to be able to do same type of file hashing correctly with tetragon.  Since the process isn't being paused as file accesses happen it will be hard to reliably take hashes without children processes potentially writing while attempting to take the hash.

attestation/commandrun/commandrun.go
colek42: done

attestation/commandrun/commandrun.go
colek42: It looks like "TRACEME" is going to be what we want to use unless we want to rewrite the exec_linux package.  https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/syscall/exec_linux.go;drc=0a52d80666ddaa557cec17ad9166e2514b0bb6d4;l=62

mikhailswift: Correct, that's what we're setting for the ptrace tracing code I linked above.

To make that work we'd have to start the process with that attribute set, setup tetragon, and then attach and detach a ptracer to the process to get it to kick off.

The only other way around it would be to handle the forking and execve'ing ourselves.

colek42: I don't want to write that fork/exec code for this feature -- seems like a can of worms.  I'll refactor your ptrace code to work with both tracers.

@mikhailswift  Do you think it would be useful to write a tracer interface?

mikhailswift: I can't say for sure but I have a hard time imagining an interface that would work well for both current tracing models we have since they differ so much.

At the moment just make the code that sets up the TRACEME syscall if either ptrace or tetragon tracing is setup and then brancing off to the correct path afterward is probably good enough.  For the ptrace code we will want to continue as normal and for the tetragon code we'll just want to detach after the first caught syscall.

attestation/commandrun/tracing_tetra.go
```diff
@@ -0,0 +1,636 @@
...
+		}
+
+		_, err = tc.client.AddTracingPolicy(tc.ctx.Context(), &tetragon.AddTracingPolicyRequest{
+			Yaml: string(j),
+		})
+		//remove sensor in case it didn't get cleaned up
+		if err != nil {
+			if strings.Contains(err.Error(), "already exists") {
+				RemoveTrace(tc.ctx.Context(), tc.client, p.Metadata.Name)
+				goto ADD_POLICY
```
mikhailswift, line 171: why a goto and label here? can't we just return the err in an if statement else fall through?

colek42: I was wondering if you were going to notice that.  It is left over after a refactor.  I'll clean it up.

mikhailswift:
Is the issue of tracing the first exec still occurring with this?

If so I may take a look into it to get this over the finish line.

PR #12, opened 2022-09-07
fix: don't return error on EOF, break instead.

colek42:
I was not able to reproduce this locally for some reason but had to set up a CI job in GitLab

Here is an example of the error:
[https://gitlab.com/testifysec/judge-platform/web/-/jobs/2994900989](https://gitlab.com/testifysec/judge-platform/web/-/jobs/2994900989)

Here is an example with the fix:
[https://gitlab.com/testifysec/judge-platform/web/-/jobs/2994887589](https://gitlab.com/testifysec/judge-platform/web/-/jobs/2994887589)

This is the binary I downloaded in the pipeline
https://github.com/testifysec/witness/releases/download/v0.0.7-test-ci/witness

### DIFF
--- a/attestation/oci/oci.go
+++ b/attestation/oci/oci.go
@@ -178,14 +178,21 @@ func (a *Attestor) parseMaifest(ctx *attestation.AttestationContext) error {
 
 	f, err := os.Open(a.tarFilePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


colek42:
This is the attestation from the successful GitLab job

image

mikhailswift (approved):

PR #16, opened 2022-09-17
RFC 3161 Timestamping Support

mikhailswift:
* Adds an array to the signatures on a dsse envelope representing
  timestamps of that signature
* Adds implementation of RFC3161 timestamps for signing and policy
  enforcement
* Adds policy support for ensuring signatures have timestamps from a
  trusted time stamp authority

### DIFF
--- a/dsse/dsse.go
+++ b/dsse/dsse.go
@@ -15,14 +15,7 @@
 package dsse
 
 import (
-	"bytes"
-	"crypto/x509"
-	"encoding/pem"
 	"fmt"
-	"io"
-	"time"
-
-	"github.com/testifysec/go-witness/cryptoutil"
 )
 
 type ErrNoSignatures struct{}
@@ -61,12 +54,20 @@ type Envelope struct {
 }
 
 type Signature struct {
-	KeyID         string   `json:"keyid"`
-	Signature     []byte   `json:"sig"`
-	Certificate   []byte   `json:"certificate,omitempty"`
-	Intermediates [][]byte `json:"intermediates,omitempty"`
+	KeyID         string               `json:"keyid"`
+	Signature     []byte               `json:"sig"`
+	Certificate   []byte               `json:"certificate,omitempty"`
+	Intermediates [][]byte             `json:"intermediates,omitempty"`
+	Timestamps    []SignatureTimestamp `json:"timestamps,omitempty"`
+}
 
-	trustedTime time.Time
+type SignatureTimestampType string
+
+const TimestampRFC3161 SignatureTimestampType = "tsp"
+
+type SignatureTimestamp struct {
+	Type SignatureTimestampType `json:"type"`
+	Data []byte                 `json:"data"`
 }
 
 // preauthEncode wraps the data to be signed or verified and it's type in the DSSE protocol's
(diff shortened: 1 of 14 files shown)


mikhailswift:
I broke up the dsse sign and verify code into their separate files and it busted the diff :/. 

I'll try and re-arrange the commits so the file split happens separately, otherwise reviewing this is gonna be a pain.

dsse/dsse.go
```diff
@@ -61,12 +54,20 @@ type Envelope struct {
...
 type Signature struct {
-	KeyID         string   `json:"keyid"`
-	Signature     []byte   `json:"sig"`
-	Certificate   []byte   `json:"certificate,omitempty"`
-	Intermediates [][]byte `json:"intermediates,omitempty"`
+	KeyID         string               `json:"keyid"`
+	Signature     []byte               `json:"sig"`
+	Certificate   []byte               `json:"certificate,omitempty"`
+	Intermediates [][]byte             `json:"intermediates,omitempty"`
+	Timestamps    []SignatureTimestamp `json:"timestamps,omitempty"`
```
colek42, line 61: This looks like a timestamp signature on the signature, correct?

dsse/dsse.go
```diff
@@ -61,12 +54,20 @@ type Envelope struct {
...
+}
+
+type SignatureTimestampType string
+
+const TimestampRFC3161 SignatureTimestampType = "tsp"
 
-	trustedTime time.Time
+type SignatureTimestamp struct {
+	Type SignatureTimestampType `json:"type"`
+	Data []byte                 `json:"data"`
```
colek42, line 70: How is this data encoded?

dsse/sign.go
```diff
@@ -0,0 +1,95 @@
...
+
+type SignOption func(*signOptions)
+
+func SignWithSigners(signers ...cryptoutil.Signer) SignOption {
+	return func(so *signOptions) {
+		so.signers = signers
+	}
+}
+
+func SignWithTimestampers(timestampers ...Timestamper) SignOption {
```
colek42, line 29: I like how you implement this as an option

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/anchore/stereoscope v0.0.0-20220708133445-777471f38c5b
 	github.com/anchore/syft v0.53.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/digitorus/timestamp v0.0.0-20220704143351-8225fba02d52
```
colek42, line 9: Eventually, I'd like this to be in tree or a fork under our control

go.mod
```diff
@@ -36,6 +37,7 @@ require (
 	github.com/containerd/containerd v1.6.6 // indirect
 	github.com/containerd/stargz-snapshotter/estargz v0.12.0 // indirect
 	github.com/coreos/go-oidc/v3 v3.2.0 // indirect
+	github.com/digitorus/pkcs7 v0.0.0-20220704143225-a9c8106cbfc6 // indirect
```
colek42, line 40: See above comment

dsse/dsse.go
mikhailswift: Correct

dsse/dsse.go
mikhailswift: Base64. Go json marshals all byte arrays to base64.

go.mod
mikhailswift: Maintaining a fork of every lib we use is a maintenance nightmare

mikhailswift: As far as what data actually is that is an implementation detail of the timestamper being used. 

In the case of TSP it’s an ASN1 encoded Timestamp Request object.

dsse/dsse.go
mikhailswift: As far as what data actually is that is an implementation detail of the timestamper being used.

In the case of TSP it’s an ASN1 encoded Timestamp Request object.

go.mod
mikhailswift: We need to be judicial about what we fork and why. 

Forking repos incurs maintenance costs and burdens.

colek42: I looked into it a bit more; the repo is maintained by a real company, https://www.digitorus.com/, based in the Netherlands that specializes in PKI.  I don't see any immediate concerns.

Initially, seeing a small repo with only two maintainers on a critical code path gives me pause.

timestamp/tsp.go
```diff
@@ -0,0 +1,160 @@
...
+func TimestampWithRequestCertificate(requestCertificate bool) TSPTimestamperOption {
+	return func(t *TSPTimestamper) {
+		t.requestCertificate = requestCertificate
+	}
+}
+
+func NewTimestamper(opts ...TSPTimestamperOption) TSPTimestamper {
+	t := TSPTimestamper{
+		hash:               crypto.SHA256,
+		requestCertificate: true,
```
colek42, line 47: Have you had any issues with the server not returning certs?  I think I remember somewhere that some TSAs do not implement this

mikhailswift: Not currently, though I've only really tested with freetsa.org currently.

We may want to add some logic to check if there is no certificate and if so reject the timestamp perhaps

colek42:
LGTM!  however, I think we need to make an issue for offline testing.

colek42 (approved):

PR #18, opened 2022-09-26
Feat/spdx sbom

colek42:
Adds SBOM creation support through https://github.com/opensbom-generator/spdx-sbom-generator.

This creates SBOMS based on the programming language package manager.  Supported languages:

```
GoMod (go)
Cargo (Rust)
Composer (PHP)
DotNet (.NET)
Maven (Java)
NPM (Node.js)
Yarn (Node.js)
PIP (Python)
Pipenv (Python)
Gems (Ruby)
Swift Package Manager (Swift)
```


attestation/sbom/sbom.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	//get files in dir
+	files, err := os.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+
+	//get the first file
+	file := files[0]
```
mikhailswift, line 101: We should probably check this array access to prevent a panic in case something weird happens and the file isn't there

mikhailswift:
One small change, otherwise looks good

mikhailswift (dismissed):

mikhailswift (approved):

PR #21, opened 2022-10-11
refactor: uses new version of archivist without grpc endpoint

mikhailswift:
Signed-off-by: Mikhail Swift <mikhail@testifysec.com>


colek42 (approved):
lgtm!

PR #22, opened 2022-10-14
feat: add github attestor

colek42:
This is ready

### DIFF
--- a/attestation/github/github.go
+++ b/attestation/github/github.go
@@ -1,0 +1,197 @@
+// Copyright 2021 The Witness Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


colek42:
This is working as expected, however, I think it may make sense to be explicit about the claims that github provides i.o.t. work with code gen tools.

colek42:
This is ready, removing the tests as the mock jwks server probably belongs as a test on the jwt package.

attestation/github/github.go
```diff
@@ -0,0 +1,197 @@
...
+	PipelineName string        `json:"pipelinename"`
+	PipelineUrl  string        `json:"pipelineurl"`
+	ProjectUrl   string        `json:"projecturl"`
+	RunnerID     string        `json:"runnerid"`
+	CIHost       string        `json:"cihost"`
+	CIServerUrl  string        `json:"ciserverurl"`
+	RunnerArch   string        `json:"runnerarch"`
+	RunnerOS     string        `json:"runneros"`
+	jwksURL      string        `json:"jwksurl"`
+	tokenURL     string        `json:"tokenurl"`
```
mikhailswift, line 75: This and the other two fields linting is yelling about will never appear in the marshaled json because they are unexported.  They should either be exported or not have json tags if they're not meant to be in the marashaled json.

mikhailswift (changes requested):

mikhailswift (approved):

PR #24, opened 2022-10-19
Adds jsonschema and typescript code generation

colek42:


mikhailswift:
I don't think the typescript definitions belong in this repository

colek42:
> 

Agree, we will need to set up a repo and npm package for them before merge

PR #31, opened 2023-02-14
fix: fixes bug where files under 512 bytes would cause EOF error

colek42:
This fixes the EOF error and some additional issues relating to small files discovered when I created some tests.

I added support for detecting pdf files to show the pattern which can be used when adding additional type detection.

### DIFF
--- a/attestation/product/product.go
+++ b/attestation/product/product.go
@@ -17,11 +17,15 @@ package product
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"mime"
+	"net/http"
+	"os"
+	"path/filepath"
+
 	"github.com/testifysec/go-witness/attestation"
 	"github.com/testifysec/go-witness/attestation/file"
 	"github.com/testifysec/go-witness/cryptoutil"
-	"net/http"
-	"os"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


attestation/product/product_test.go
```diff
@@ -15,7 +15,12 @@
 package product
 
 import (
+	"archive/tar"
 	"crypto"
+	"io/ioutil"
```
mikhailswift, line 20: io/ioutil is deprecated

attestation/product/product.go
```diff
@@ -126,15 +132,69 @@ func (a *Attestor) Subjects() map[string]cryptoutil.DigestSet {
 }
 
 func getFileContentType(file *os.File) (string, error) {
-	// Only the first 512 bytes are used to sniff the content type.
-	buffer := make([]byte, 512)
-	_, err := file.Read(buffer)
```
mikhailswift, line 131: This code was fine, why not just check for EOF and not return in that case?

attestation/product/product.go
```diff
@@ -126,15 +132,69 @@ func (a *Attestor) Subjects() map[string]cryptoutil.DigestSet {
...
 	}
+	fileSize := fileInfo.Size()
 
-	// Use the net/http package's handy DectectContentType function. Always returns a valid
-	// content-type by returning "application/octet-stream" if no others seemed to match.
+	// Read up to 512 bytes from the file.
+	bufferSize := int(math.Min(float64(fileSize), 512))
+	buffer := make([]byte, bufferSize)
+	_, err = file.Read(buffer)
+	if err != nil && err != io.EOF {
```
mikhailswift, line 146: As above, you could just use file.Read and check for EOF as you are here.

mikhailswift (changes requested):

mikhailswift:
There's a bit more changes here than just fixing the EOF error not being handled when the file is less than 512 bytes, should this be separated into a separate PR or can you explain a bit about what spawned the additional changes?

colek42:
When I added the tests, I found some edge cases where file content detection was not done correctly with the HTTP package. especially when the file was under 512 bytes

mikhailswift (approved):

PR #32, opened 2023-02-16
feat: attestor config options

mikhailswift:
This adds the ability for attestors to expose available configuration
options including their names, descriptions, and a function that can be
used to set the configuration option on the instantiated attestor.

This will let users of the registry to expose options to the end user,
such as through CLI flags in the case of Witness CLI

### DIFF
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -24,21 +24,23 @@ import (
 	"github.com/testifysec/go-witness/log"
 )
 
-type ErrInvalidOption struct {
-	Option string
-	Reason string
-}
-
 type RunType string
 
 const (
-	Internal    RunType = "internal"
-	PreRunType  RunType = "pre"
-	PostRunType RunType = "post"
+	PreMaterialRunType RunType = "prematerial"
+	MaterialRunType    RunType = "material"
+	ExecuteRunType     RunType = "execute"
+	ProductRunType     RunType = "product"
+	PostProductRunType RunType = "postproduct"
 )
 
 func (r RunType) String() string {
 	return string(r)
+}
+
+type ErrInvalidOption struct {
+	Option string
+	Reason string
 }
 
 func (e ErrInvalidOption) Error() string {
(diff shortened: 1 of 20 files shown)


attestation/product/product.go
```diff
@@ -45,12 +45,45 @@ var (
...
-	})
+	}, attestation.StringConfigOption("includeglob", "Glob that will be used to match files to record as products", func(a attestation.Attestor, includeGlob string) (attestation.Attestor, error) {
+		attestor, ok := a.(*Attestor)
+		if !ok {
+			return nil, fmt.Errorf("unexpected attestation type %T", a)
+		}
+
+		attestor.includeGlob = includeGlob
+		return attestor, nil
+	}), attestation.StringConfigOption("excludeglob", "Glob that will be used to match files to exclude as products", func(a attestation.Attestor, excludeGlob string) (attestation.Attestor, error) {
```
colek42, line 56: Im thinking should take priority.  Also, this will only be for subject creation when implemented, right?

mikhailswift: To be honest I haven't put any thought into the actual implementation of these actual settings.  It depends on the user case that's driving these options but my thought is yeah, exclude would take precedence.

Currently these two options are not implemented and that will likely belong in it's own issue, this PR is mostly focused around generally supporting configuring attestors through the CLI.

attestation/context.go
```diff
@@ -32,9 +32,11 @@ type ErrInvalidOption struct {
 type RunType string
 
 const (
-	Internal    RunType = "internal"
```
mikhailswift, line 35: Internal being a special run type always kind of felt bad, imo, and having 3 attestors treated specially made for hard to reason about edge cases when configuring these attestors through the CLI.

The special edge cases around these 3 attestors I believe are covered by creating the material and product run stages, though they honestly probably could just live inthe pre and post stages. The other things that made these three special are somewhat witness CLI specific, as other use cases of go-witness may not be concerned about files or commands at all (consider approval attestations, etc), so those cases have been moved to the CLI code base.

colek42 (dismissed):
This all looks good to me.  What needs to get done for us to get this merged in?

mikhailswift:
This should be ready for merge now.

A greater doc rework needs to be made in the near future.

mikhailswift:
Actually hold up on merge for a few

mikhailswift:
Okay, this is ready now.

colek42 (approved):

PR #36, opened 2023-02-21
feat: add support for github id token

colek42:
This is a major refactor of the Fulcio code along with a bunch of tests to support getting certs from Fulcio for a GitHub OIDC token

fixes: https://github.com/testifysec/witness/issues/248


signer/fulcio/github.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return "", err
+	}
+	req.Header.Add("Authorization", "bearer "+bearer)
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	body, err := readResponseBody(resp.Body)
```
mikhailswift, line 43: why not just use io.readall here

signer/fulcio/fulcio.go
```diff
@@ -44,104 +53,200 @@ func Signer(ctx context.Context, funcioURL string, oidcIssuer string, oidcClient
...
-		return nil, err
-	}
+	var raw string
 
-	certResp, err := getCert(fulcioSigner.(*signature.RSAPKCS1v15Signer), fClient, oidcIssuer, oidcClientID)
-	if err != nil {
-		return nil, err
-	}
+	switch {
+	// if we're running in GitHub Actions
```
mikhailswift, line 59: is this an incomplete comment? what's going on if we are running in github actions?

signer/fulcio/fulcio.go
```diff
@@ -20,21 +20,30 @@ import (
...
 	"github.com/testifysec/go-witness/cryptoutil"
+	"github.com/testifysec/go-witness/log"
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
-func Signer(ctx context.Context, funcioURL string, oidcIssuer string, oidcClientID string) (cryptoutil.Signer, error) {
-	fClient, err := newClient(funcioURL)
+func Signer(ctx context.Context, funcioURL string, oidcIssuer string, oidcClientID string, tokenOption string) (cryptoutil.Signer, error) {
+
+	fClient, err := NewClient(funcioURL)
```
mikhailswift, line 46: typo -- funcio vs fulcio

signer/fulcio/github.go
```diff
@@ -0,0 +1,73 @@
...
+	client := &http.Client{}
+
+	//add audient "&audience=witness" to the end of the tokenURL, parse it, and then add it to the query
+	u, err := url.Parse(tokenURL)
+	if err != nil {
+		return "", err
+	}
+
+	q := u.Query()
+	q.Add("audience", audience)
```
mikhailswift, line 28: if audience already exists in the tokenUrl passed to this function `q.Add` will add multiple audiences to the query string. is that desired? should we check if it exists and replace it if it does instead?

mikhailswift (commented):
just a quick glance over some stuff, will take a closer look tomorrow

signer/fulcio/fulcio.go
```diff
@@ -15,133 +15,286 @@
...
 	if err != nil {
 		return nil, err
 	}
+
 	return witnessSigner, nil
 }
 
-func getCert(signer *signature.RSAPKCS1v15Signer, fc api.Client, oidcIssuer string, oidcClientID string) (*api.CertificateResponse, error) {
-	tok, err := oauthflow.OIDConnect(oidcIssuer, oidcClientID, "", "", oauthflow.DefaultIDTokenGetter)
+func GetCert(ctx context.Context, key *rsa.PrivateKey, fc fulciopb.CAClient, token string) (*fulciopb.SigningCertificate, error) {
```
mikhailswift, line 176: Is GetCert called from outside this package?  If not I think we should leave it unexported

signer/fulcio/fulcio.go
```diff
@@ -15,133 +15,286 @@
...
+			return nil, err
+		}
+
+	case tokenOption != "":
+		raw = tokenOption
 
-	if block.Type != "CERTIFICATE" {
-		return nil, fmt.Errorf("failed to parse certificate PEM")
+	case tokenOption == "" && isatty.IsTerminal(os.Stdin.Fd()):
+		tok, err := oauthflow.OIDConnect(oidcIssuer, oidcClientID, "", "", oauthflow.DefaultIDTokenGetter)
```
mikhailswift, line 112: I think we'll eventually need to find a way to break this out of the library.  I don't think anything that will take over user interaction belongs in library code.  This is fine for now but something to mull over.

mikhailswift (commented):
One question on the API, otherwise this looks good

signer/fulcio/fulcio.go
colek42: good point

signer/fulcio/fulcio.go
colek42: I'm open to ideas

mikhailswift:
Can we get the change to GetCert and this branch updated so it can be merged @colek42

colek42:
@mikhailswift this should be ready to go

mikhailswift:
@colek42 one more set of merge conflicts, sorry!

mikhailswift (approved):

PR #39, opened 2023-02-22
Add additional git info

colek42:


kriscoleman (dismissed):
I'm still learning this side of the codebase, but I love these changes. Lots of new git meta!

attestation/git/git.go
```diff
@@ -113,19 +219,57 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
...
+	ds, err = cryptoutil.CalculateDigestSetFromBytes([]byte(a.CommitterEmail), a.hashes)
+	if err != nil {
+		return nil
+	}
+
+	subjects[subjectName] = ds
+
+	//add parent hashes
+	for _, parentHash := range a.ParentHashes {
+		subjectName = fmt.Sprintf("parenthash:%v", parentHash)
```
colek42, line 252: we might not want this as a subject.  Do we want to pull up children when we search for parents?  We can always filter on name, but that adds complications.

attestation/git/git.go
```diff
@@ -85,7 +112,86 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
...
+	commit, err := repo.CommitObject(head.Hash())
+	if err != nil {
+		return err
+	}
+
+	a.CommitDigest = cryptoutil.DigestSet{
+		{
+			Hash:   crypto.SHA1,
+			GitOID: false,
+		}: commit.Hash.String(),
```
mikhailswift, line 124: I'm not sure that a digest set makes sense used like this. The resulting data in this case is going to look something like

```
"commitdigest" {
  "63ff2342b54bd1ffc9dbdf527130e7dcab51d8d1": {
    "sha1": "63ff2342b54bd1ffc9dbdf527130e7dcab51d8d1"
  }
}
```

This should probably just look like

```
"commitdigest": {
  "sha1": "63ff2342b54bd1ffc9dbdf527130e7dcab51d8d1"
}
```

attestation/git/git.go
```diff
@@ -49,8 +49,19 @@ type Status struct {
...
-	Status     map[string]Status `json:"status,omitempty"`
+	CommitHash     string               `json:"commithash"`
+	Author         string               `json:"author"`
+	AuthorEmail    string               `json:"authoremail"`
+	CommitterName  string               `json:"committername"`
+	CommitterEmail string               `json:"committeremail"`
+	CommitDate     string               `json:"commitdate"`
+	CommitMessage  string               `json:"commitmessage"`
+	Status         map[string]Status    `json:"status,omitempty"`
+	CommitDigest   cryptoutil.DigestSet `json:"commitdigest,omitempty"`
```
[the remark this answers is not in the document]
mikhailswift: I'd be inclined to just bump this attestor to v0.2 and replace `commithash` with this, as most people are familiar with the terminology 'commit hash' versus 'commit digest'

Also, is it possible currently in git for a repo to have both sha-1 and sha-256 commit hashes? Currently I believe the entire repo has to use one or the other.

attestation/git/git.go
mikhailswift: We should be in the habit of always filtering by name as well as digest when searching archivista for subjects.

The only use case of having the parents as subjects here I can think of is to find children of a commit we're currently looking at, since we already have the parents on the attestor itself.  That could have some value, potentially.

mikhailswift:
Only thing that needs to change before this gets merged are the json tags for taggerName and taggerEmail. We can leave `commitdigest` as-is for now, though I think there's value in just having it be the canonical identifier of the repo versus multiple hashes.

Parents being in the subjects I'll leave up to you. I can see some value having them there, and while it will make searching Archivista with only a digest a bit more complicated we really should only be searching archivista with a combination of subject name and digest, really.

attestation/git/git.go
```diff
@@ -85,7 +112,86 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
...
+	var tagList []Tag
+
+	err = tags.ForEach(func(t *object.Tag) error {
+
+		//check if the tag points to the head
+		if t.Target.String() != head.Hash().String() {
+			return nil
+		}
+
+		fmt.Printf("tag: %s\n", t.Name)
```
mikhailswift, line 176: Remove

mikhailswift:
Can we get the few changes done so this can be merged @colek42 ?

attestation/git/git.go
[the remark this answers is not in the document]
colek42: We can calculate both.  However, tt wasn't clear that go-git v5 supported it and may be something we need to do manually

colek42:
@mikhailswift this should be ready

mikhailswift (approved):

PR #41, opened 2023-02-26
feat: add include and exclude glob options to product attestor

mikhailswift:
This allows users to configure the product attestor to only include certain files from the product attestor as subjects to the attestation.

### DIFF
--- a/attestation/product/product.go
+++ b/attestation/product/product.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


kriscoleman (dismissed):
seems like a nice way to add glob include/excludes, still learning this part of codebase.

colek42 (changes requested):

colek42:
sorry, I apparently cant read

colek42 (approved):

PR #42, opened 2023-02-27
feat: attestation timestamps

colek42:


attestation/collection.go
```diff
@@ -29,27 +30,34 @@ type Collection struct {
 }
 
 type CollectionAttestation struct {
-	Type        string   `json:"type"`
-	Attestation Attestor `json:"attestation"`
+	Type        string    `json:"type"`
+	Attestation Attestor  `json:"attestation"`
+	StartTime   time.Time `json:"start_time"`
+	EndTime     time.Time `json:"end_time"`
```
mikhailswift, line 36: i don't believe we use underscores in any other json tags, so these should probably be `starttime` and `endtime` for consistency

mikhailswift (commented):
minor thing with the json tags. this looks great!

attestation/collection.go
colek42: thank you

mikhailswift (approved):

PR #43, opened 2023-02-28
[WIP] refactor tracing and add network tracing

colek42:


attestation/commandrun/commandrun.go
```diff
@@ -141,40 +156,93 @@ func (rc *CommandRun) RunType() attestation.RunType {
...
 	stdoutWriter := io.MultiWriter(stdoutWriters...)
 	stderrWriter := io.MultiWriter(stderrWriters...)
+
+	// Set the command's stdout and stderr to the multi-writers
 	c.Stdout = stdoutWriter
 	c.Stderr = stderrWriter
+
+	// If tracing is enabled, set the SysProcAttr to enable PTRACE
 	if r.enableTracing {
-		enableTracing(c)
```
mikhailswift, line 160: this needs to be left as-is, SysProcAttr.Ptrace is linux specific.

attestation/commandrun/commandrun.go
```diff
@@ -141,40 +156,93 @@ func (rc *CommandRun) RunType() attestation.RunType {
...
+	}
+
+	c.SysProcAttr.Credential = &syscall.Credential{
+		Uid: uint32(os.Geteuid()),
+		Gid: uint32(os.Getegid()),
 	}
 
 	var err error
+
+	// Start the command
```
mikhailswift, line 223: These types of comments arent adding a ton of value, `c.Start` is fairly self documenting

attestation/commandrun/tracing_linux.go
```diff
@@ -46,12 +50,6 @@ type ptraceContext struct {
 	environmentBlockList map[string]struct{}
 }
 
-func enableTracing(c *exec.Cmd) {
```
mikhailswift, line 49: as above, enabling tracing is fairly platform dependent -- linux is the only platform that supports ptrace so this needs to be in the linux gated file

attestation/commandrun/tracing_linux.go
```diff
@@ -74,54 +73,104 @@ func (r *CommandRun) trace(c *exec.Cmd, actx *attestation.AttestationContext) ([
 	return pctx.procInfoArray(), nil
 }
 
-func (p *ptraceContext) runTrace() error {
+func (p *ptraceContext) runTrace(ctx context.Context) error {
+	// Create a new channel for reporting errors.
+	errChan := make(chan error)
```
mikhailswift, line 78: this is all synchronous code, we don't really need a channel and go routine to print logging messages here, calling log.Errorf at the erros site is easier and adds less cognitive burden when reasoning about what's going on with these errors.

attestation/commandrun/tracing_linux.go
```diff
@@ -292,18 +536,47 @@ func getPPIDFromStatus(status []byte) (int, error) {
...
-			isVuln := strings.TrimSpace(parts[1])
-			if strings.Contains(isVuln, "vulnerable") {
-				return true
-			}
+func (p *ptraceContext) readSyscallData(pid int, addr uintptr, size int) ([]byte, error) {
+	buf := make([]byte, size)
+	n := 0
+	for n < size {
+		word := make([]byte, WORD_SIZE)
+		_, err := unix.PtracePeekData(pid, uintptr(addr)+uintptr(n), word)
```
mikhailswift, line 544: ptracepeekdata is very slow -- that's why we use ProcessVMReadv when reading process data elsewhere.  is that possible here?

attestation/commandrun/tracing_linux.go
```diff
@@ -142,84 +191,279 @@ func (p *ptraceContext) nextSyscall(pid int) error {
 	return nil
 }
 
-func (p *ptraceContext) handleSyscall(pid int, regs unix.PtraceRegs) error {
-	argArray := getSyscallArgs(regs)
-	syscallId := getSyscallId(regs)
+func (p *ptraceContext) handleExeCve(pid int, argArray []uintptr) error {
```
mikhailswift, line 194: breaking these out into their own functions is very welcome change. I don't think the `C` in `ExeCve` should be capitalized, though

PR #45, opened 2023-05-21
WIP: adding support for omnibor filesystem attestation

fkautz:
Signed-off-by: Frederick Kautz <fkautz@alumni.cmu.edu>


attestation/file/file.go
```diff
@@ -30,11 +33,16 @@ import (
 // returned map of artifacts.
 func RecordArtifacts(basePath string, baseArtifacts map[string]cryptoutil.DigestSet, hashes []crypto.Hash, visitedSymlinks map[string]struct{}) (map[string]cryptoutil.DigestSet, error) {
 	artifacts := make(map[string]cryptoutil.DigestSet)
+	trail := omnitrail.New(omnitrail.WithSha1(), omnitrail.WithSha256())
```
colek42, line 36: We should be able to pass in an exclude pattern, for example do we want to include the .git directory?

colek42, line 36: I think we want support for other hashers, such as normal sha256 so we can make correlations with online databases such as deps.dev, 

ref:  https://github.com/google/deps.dev/blob/main/examples/go/artifact_query/main.go

colek42 (commented):
This looks great and exactly what we want.  We have a couple of features we need for this initial use case.  See the inline comments

colek42:
this build requires updating the omnibor module in witness for the build to work.

attestation/file/file.go
fkautz: That's a good idea, we could accept a regex pattern easily.

colek42: we use a glob pattern in products:  https://github.com/testifysec/go-witness/blob/main/attestation/product/product.go#L49

fkautz: That's a good idea. Golang had some hardcoded hashers. We could accept such a list and generate the attestations as required.

We should also change the current implementation. It reads the file twice to generate both sha1 and sha256 gitoids but this should be reimplemented to support a multiplexed reader especially when this is implemented.

ChaosInTheCRD:
Hi everyone 👋 

Just wondering what the current state of this PR is. I am just trying to clean up the PRs on the repository, and I was wondering whether progress could be tracked in an Issue, and we can reopen this PR when it is ready for final review?

Let me know what you think,

colek42:
@ChaosInTheCRD feel free to close the PR, but we should retain the branch.  We are a ways out from this being ready

PR #52, opened 2023-08-07
feat: vault signer

mikhailswift:
* Adds support for using Vault's PKI secrets engine as a signer provider
* Adds a generic registry concept used to expose configuration options for users of the library
* Uses the registry for attestors and signers


kriscoleman (approved):
looking at this second time and it lgtm 

appreciate the refactoring and new tests along with the new feat

jkjell (approved):

PR #53, opened 2023-09-19
Add support for controller-gen deepcopy

jkjell:
Add support for `controller-gen` deepcopy of policy package structs. This will be required for archivista data provider controllers.


policy/policy.go
```diff
@@ -24,23 +24,28 @@ import (
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const PolicyPredicate = "https://witness.testifysec.com/policy/v0.1"
 
+// +kubebuilder:object:generate=true
 type Policy struct {
-	Expires              time.Time            `json:"expires"`
+	Expires              metav1.Time          `json:"expires"`
```
mikhailswift, line 35: Does this affect the json marshalling/unmarshalling of this struct at all?

jkjell: Possibly? If it does, for the better I think. But, we probably won't notice. From the Kubebuilder book:
> There’s one other special type that we use: metav1.Time. This functions identically to time.Time, except that it has a fixed, portable serialization format.

mikhailswift:
Ran witness's tests with these changes and the expiry date all works, so this lgtm!

mikhailswift (approved):

PR #55, opened 2023-10-16
verification attestor

mikhailswift:
Fixes https://github.com/testifysec/witness/issues/14

The rough idea here is to have `go-witness.Verify` actually be a `go-witness.Run` with a verification attestor as it's `Execute` attestor. This allows a few different things:

1. We can create verification attestations. We're currently using the SLSA VSA format. There is some chatter about moving this under the in-toto attestation group: https://github.com/in-toto/attestation/issues/277 . This would involve generalizing some of the SLSA specific data in the VSA, which would be ideal.
2. We can collect other attestations about the system that verified an artifact. Since the policy verification is happening within the verification attestor in an attestation context, we can also run our other suite of attestors, such as the GitHub or GitLab attestors, and collect more information about where an artifact was verified.
3. We can use verification attestations in a handful of ways, including "caching" decisions within an admission controller. A policy could potentially define a way to act as a quick pass if we see we've already verified an artifact against a specified policy. This could be time-gated, so we must re-evaluate if the verification attestation is past a certain age.
4. Act as a summary of a group of raw attestations. Some attestations may not be fit for public consumption. In these cases, organizations could publish the VSA instead of raw attestations to their end users.

There are a few philosophical questions and changes to the go-witness API happening here we should think a bit about:

In main today, the DSSE signing code will throw up if at least one signer isn't provided. This makes sense; why wrap something in a DSSE envelope if it won't be signed? However, we don't want every verifier to have keys distributed to them.  Users should be able to run `go-witness.Verify` without providing a key. If one is provided, a signed VSA should be created. In this draft changeset, this error is removed. An unsigned envelope will be returned if no signers are provided to the DSSE code.

This has the side-effect of allowing users to run `go-witness.Run` without a signer provided. This may not be a bad thing, we would hand back an unsigned envelope of attestations. This could allow users to play with Witness more quickly and sort out key distribution once ready.

While we offer no guarantees of backward compatibility, this may involve some breaking changes to `go-witness`'s surface APIs. We shouldn't let this stop us from making choices, but we should be cognizant of the pain this will cause and ensure we do it for the right reasons.


attestation/verification/verification.go
```diff
@@ -0,0 +1,209 @@
+package verification
+
```
naveensrinivasan, line 2: A suggestion would be to use generics to accommodate other types in the future. 

TBH, I don't know every bit and I could be totally wrong. 

``` go
const (
	Name = "policyverify"
	Type = slsa.VerificationSummaryPredicate
)

var (
	_ attestation.Subjecter = &Attestor[cryptoutil.Verifier]{}
	_ attestation.Attestor  = &Attestor[cryptoutil.Verifier]{}
)

type Verifier interface {
	Verify(body io.Reader, signature []byte) error
}

type Attestor[T Verifier] struct {
	slsa.VerificationSummary

	policyEnvelope   dsse.Envelope
	policyVerifiers  []T
	collectionSource source.Sourcer
	subjectDigests   []string
}

type Option[T Verifier] func(*Attestor[T])

func VerifyWithPolicyVerifiers[T Verifier](policyVerifiers []T) Option[T] {
	return func(vo *Attestor[T]) {
		vo.policyVerifiers = append(vo.policyVerifiers, policyVerifiers...)
	}
}

func VerifyWithSubjectDigests[T Verifier](subjectDigests []cryptoutil.DigestSet) Option[T] {
	return func(vo *Attestor[T]) {
		for _, set := range subjectDigests {
			for _, digest := range set {
				vo.subjectDigests = append(vo.subjectDigests, digest)
			}
		}
	}
}

func VerifyWithCollectionSource[T Verifier](source source.Sourcer) Option[T] {
	return func(vo *Attestor[T]) {
		vo.collectionSource = source
	}
}

func New[T Verifier](opts ...Option[T]) *Attestor[T] {
	a := &Attestor[T]{}
	for _, opt := range opts {
		opt(a)
	}

	return a
}

func (vs *Attestor[T]) Name() string {
	return Name
}

func (vs *Attestor[T]) Type() string {
	return Type
}

func (vs *Attestor[T]) RunType() attestation.RunType {
	return attestation.ExecuteRunType
}

func (vs *Attestor[T]) Subjects() map[string]cryptoutil.DigestSet {
	panic("not implemented") // TODO: Implement
}
```

attestation/verification/verification.go
```diff
@@ -0,0 +1,209 @@
...
+	trustBundlesById, err := pol.TrustBundles()
+	if err != nil {
+		return fmt.Errorf("failed to load policy trust bundles: %w", err)
+	}
+
+	roots := make([]*x509.Certificate, 0)
+	intermediates := make([]*x509.Certificate, 0)
+	for _, trustBundle := range trustBundlesById {
+		roots = append(roots, trustBundle.Root)
+		intermediates = append(intermediates, intermediates...)
```
naveensrinivasan, line 125: In the Attest method, there's a line where the intermediates slice is appended to itself:

This operation duplicates the intermediates slice each time it's run. If intermediates is not empty, this could lead to unexpected behavior or bugs as the slice would grow rapidly. Here's a simple example to illustrate this:
```go
func main() {
	slice := []int{1, 2, 3}
	for i := 0; i < 3; i++ {
		slice = append(slice, slice...)
	}
	fmt.Println(slice) // Output: [1 2 3 1 2 3 1 2 3 1 2 3 1 2 3 1 2 3 1 2 3]
}
```

In this example, the slice grows rapidly due to the append operation in the loop. A similar issue could occur in your code if intermediates is not empty.

If intermediates is always empty at the time of the append operation (as it seems to be in your code), it would have no effect, but it's still potentially confusing to readers of the code.

Could it instead be this?
` intermediates = append(intermediates, trustBundle.Intermediates...)`

naveensrinivasan (commented):
A couple of suggestions.

attestation/verification/verification.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	pol := policy.Policy{}
+	if err := json.Unmarshal(vo.policyEnvelope.Payload, &pol); err != nil {
+		return fmt.Errorf("failed to unmarshal policy from envelope: %w", err)
+	}
+
+	pubKeysById, err := pol.PublicKeyVerifiers()
+	if err != nil {
+		return fmt.Errorf("failed to get pulic keys from policy: %w", err)
```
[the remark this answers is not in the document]
mikhailswift: Ooops, thanks for the catch on this.

attestation/verification/verification.go
mikhailswift: Thanks for the catch on this -- this bug existed in our previous verify function which is especially scary. We would have accidentally ignored policy-defined intermediates with this bug. 🤦

attestation/verification/verification.go
mikhailswift: Thanks for the suggestion. I don't think we have a current use case to support some other form of verifier outside of a `cryptoutil.Verifier` here currently, and we currently need our verifiers to implement the Verifier interface.

mikhailswift:
Additional changes pushed to this:

* Added the `insecure` flag to `go-witness.Run` to allow running without signers provided.
* If an ErrPolicyDenied error is received from `policy.Verify`, it is returned to the caller. This is the canonical way to communicate verification failures to callers currently, so this is an effort to maintain that consistency.
* Added signer options to `VerifyOption`s. This is a bit confusing and I don't love it... Because we also have a `RunWithRunOptions` that will forward Run options, the user could supply signers in various ways currently.

This still needs tests, but I am able to run our witness test scripts with these changes with this branch on witness: https://github.com/testifysec/witness/tree/feat/verification-attestation

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,206 @@
...
+		dsse.VerifyWithRoots(roots...),
+		dsse.VerifyWithIntermediates(intermediates...),
+		dsse.VerifyWithTimestampVerifiers(timestampVerifiers...),
+	)
+
+	accepted := true
+	policyResult, policyErr := pol.Verify(ctx.Context(), policy.WithSubjectDigests(vo.subjectDigests), policy.WithVerifiedSource(verifiedSource))
+	if _, ok := policyErr.(policy.ErrPolicyDenied); ok {
+		accepted = false
+	} else if err != nil {
```
naveensrinivasan, line 150: I think the code should check for `policyErr` instead of `err` here.

``` go
var accepted bool
	policyResult, policyErr := pol.Verify(ctx.Context(), policy.WithSubjectDigests(vo.subjectDigests), policy.WithVerifiedSource(verifiedSource))
	if _, ok := policyErr.(policy.ErrPolicyDenied); ok {
		accepted = false
	} else if policyErr != nil {
		return fmt.Errorf("failed to verify policy: %w", policyErr)
	} else {
		accepted = true
	}
```

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (vs *Attestor) RunType() attestation.RunType {
+	return attestation.ExecuteRunType
+}
+
+func (vs *Attestor) Subjects() map[string]cryptoutil.DigestSet {
+	return map[string]cryptoutil.DigestSet{}
+}
+
+func (vo *Attestor) Attest(ctx *attestation.AttestationContext) error {
```
naveensrinivasan, line 94: A nit `vo`. Having consistent receiver name with `vs` helps :)

attestation/policyverify/policyverify.go
mikhailswift: Yup, good catch

attestation/policyverify/policyverify.go
mikhailswift: Agreed, the receiver name is a byproduct of the code moving where the variable was originally of type `verifyOptions`, this should be renamed to be consistent with our new receiver.

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,206 @@
...
+	verificationResult := slsa.FailedVerificationResult
+	if accepted {
+		verificationResult = slsa.PassedVerificationResult
+	}
+
+	return slsa.VerificationSummary{
+		Verifier: slsa.Verifier{
+			ID: "witness",
+		},
+		TimeVerified: time.Now(),
```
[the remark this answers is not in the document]
mikhailswift: It shouldn't matter too much, since when this gets marshaled to JSON it will have the time zone offset included. But being consistent with times and having them be UTC wouldn't be a bad thing.

mikhailswift:
I added a second commit to this to support some work I'm doing in a separate project that's using go-witness. It's not directly related to VSAs, but I needed it for work in the archivista-data-provider.

When this PR gets moved out of draft and is ready for merge, the second commit will be broken out into it's own PR as it is a sizeable change in it's own right.

naveensrinivasan:
> I added a second commit to this to support some work I'm doing in a separate project that's using go-witness. It's not directly related to VSAs, but I needed it for work in the archivista-data-provider.
> 
> When this PR gets moved out of draft and is ready for merge, the second commit will be broken out into it's own PR as it is a sizeable change in it's own right.

Currently, the cryptoutil.DigestValue type is used extensively across the codebase. While this works well for our current needs, it might make the code more rigid and harder to test.

One way to improve this could be to define an interface that cryptoutil.DigestValue satisfies. This would allow us to create mock implementations for testing different scenarios and make the code more flexible. Here's a simple example of what this might look like:

``` golang
type Digest interface {
    New() hash.Hash
    // other methods...
}

type DigestValue struct {
    // fields...
}

func (d DigestValue) New() hash.Hash {
    // implementation...
}

func SomeFunction(d Digest) {
    hash := d.New()
    // use hash...
}
```

In this example, SomeFunction can now accept any object that satisfies the Digest interface. This makes it easier to test SomeFunction with different implementations of Digest.

This is just a suggestion.

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,223 @@
...
+		dsse.VerifyWithIntermediates(intermediates...),
+		dsse.VerifyWithTimestampVerifiers(timestampVerifiers...),
+	)
+
+	accepted := true
+	policyResult, policyErr := pol.Verify(ctx.Context(), policy.WithSubjectDigests(a.subjectDigests), policy.WithVerifiedSource(verifiedSource))
+	if _, ok := policyErr.(policy.ErrPolicyDenied); ok {
+		accepted = false
+	} else if policyErr != nil {
+		return fmt.Errorf("failed to verify policy: %w", err)
```
BHunter2889, line 173: This was accidentally missed when changing to the check for `policyErr`:
```suggestion
		return fmt.Errorf("failed to verify policy: %w", policyErr)
```

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,223 @@
...
+	Name = "policyverify"
+	Type = slsa.VerificationSummaryPredicate
+)
+
+var (
+	_ attestation.Subjecter = &Attestor{}
+	_ attestation.Attestor  = &Attestor{}
+)
+
+type Attestor struct {
```
BHunter2889, line 44: Does this need to be registered in an init() like the others? I know the `RunType` `ExecuteRunType` already is registered, but they are registered to different Attestor Types. 

This would be missing from `FactoryByType` for example if I'm not mistaken... is it expected to be accessed in this way?

ChaosInTheCRD: I am right in saying that https://github.com/in-toto/go-witness/pull/55/files#diff-f12a0e5e9e75cb021efdaf70898369096fe67b0aa85fc890b709c54e9183e9d3R48 satisfies this @mikhailswift ? If so we can go ahead and resolve this comment. Thanks @BHunter2889 😄

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,338 @@
...
+func init() {
+	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor {
+		return New()
+	})
+}
+
+type Attestor struct {
+	slsa.VerificationSummary
+
+	policyTimestampAuthorities []timestamp.TimestampVerifier
```
ChaosInTheCRD, line 56: Just being sure to ask what springs to mind, do we need all of these extra fields in the attestor? I can imagine that this information can be of use when performing validation on what was verified and how, but I supose this comes at the consequence of bloating the predicate output? Isn't one of the benefits of a VSA that it is lightweight and provides minimal predicate inspection by the final verifier? @mikhailswift

ChaosInTheCRD: Oh ignore me here. I see that these are not fields to be added to the predicate but instead supporting fields for attestor functionality. The `slsa.VerificationSummary` is the only field that is used as the predicate.

One thing that springs to mind is that when we have the time, better moulding the structures so that predicates are defined very clearly could be a good thing 😄

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,338 @@
...
+		verificationResult = slsa.PassedVerificationResult
+	}
+
+	return slsa.VerificationSummary{
+		Verifier: slsa.Verifier{
+			ID: "witness",
+		},
+		TimeVerified: time.Now(),
+		Policy: slsa.ResourceDescriptor{
+			URI:    policyDigest[cryptoutil.DigestValue{Hash: crypto.SHA256, GitOID: false}], //TODO: find a better value for this...
```
ChaosInTheCRD, line 273: just picking up on the TODO here. Given that we can't provide a public location of the policies, wouldn't we be better (for now) using `https://witness.testifysec.com/policy/v0.1` for this URI field?

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,338 @@
...
+			Digest: policyDigest,
+		},
+		InputAttestations:  inputAttestations,
+		VerificationResult: verificationResult,
+	}, nil
+}
+
+func verifyPolicySignature(ctx context.Context, vo *Attestor) error {
```
ChaosInTheCRD, line 281: Noticed that this is a copy of https://github.com/in-toto/go-witness/blob/main/verify.go#L182. Would we not be better exporting the original function? alternatively we could stick it in an `/internal` location so that both `verify.go` and `policyverify.go` can make use of it without the copying? Unless there's something I am missing and this function is slightly different to the one in `verify.go`

run.go
```diff
@@ -31,42 +29,65 @@ import (
...
+	signers         []cryptoutil.Signer
 	attestors       []attestation.Attestor
 	attestationOpts []attestation.AttestationContextOption
 	timestampers    []timestamp.Timestamper
+	insecure        bool
 }
 
 type RunOption func(ro *runOptions)
 
+// RunWithInsecure will allow attestations to be generated unsigned. If insecure is true, RunResult will not
```
ChaosInTheCRD, line 41: I think that while I am cautious that we need to introduce `insecure` with a high level of caution (really discourage users from doing it), `--insecure` is fine in my eyes. Just wondering what we can do to encourage signing over insecure further if we're going to introduce it. Maybe an extra log warning to specify that they are running as such and it is not an encouraged pattern to introduce.

attestation/policyverify/policyverify.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/in-toto/go-witness/policy"
+	"github.com/in-toto/go-witness/slsa"
+	"github.com/in-toto/go-witness/source"
+	"github.com/in-toto/go-witness/timestamp"
+)
+
+const (
+	Name    = "policyverify"
+	Type    = slsa.VerificationSummaryPredicate
+	RunType = attestation.ExecuteRunType
```
ChaosInTheCRD, line 39: I don't wish to make life more difficult in the process of getting this merged, but I think a new type should be declared here in to define its type.

I just ran the code locally, and in its current form, there's nothing from stopping me from trying to run this attestor during `witness run`. We should have an easy way to distinguish this attestor from the others that exist, and to me a `verify` type is the way we should to it.

Of course this is a small thing to implement (a variable declaration and a line somewhere in the `Run` logic to ensure that any `verify` attestors can't be executed, but I think it's crucial for user experience.

ChaosInTheCRD: This is addressed in a PR I created [here](https://github.com/in-toto/go-witness/pull/230)

run.go
ChaosInTheCRD: This is a nit so no action is required 👍

attestation/policyverify/policyverify.go
ChaosInTheCRD: This is a nit so no action is required 👍

attestation/policyverify/policyverify.go
ChaosInTheCRD: This is addressed in a PR I created https://github.com/in-toto/go-witness/pull/230

attestation/policyverify/policyverify.go
ChaosInTheCRD: This is addressed in a PR I created https://github.com/in-toto/go-witness/pull/230

jkjell (approved):

PR #58, opened 2023-10-25
Changed to pointer receiver  when both were mixed

naveensrinivasan:
- Changed to pointer receiver when both were mixed.
- https://go.dev/doc/faq#methods_on_values_or_pointers


cryptoutil/digestset.go
```diff
@@ -114,9 +114,9 @@ func (first DigestSet) Equal(second DigestSet) bool {
 	return hasMatchingDigest
 }
 
-func (ds DigestSet) ToNameMap() (map[string]string, error) {
+func (first DigestSet) ToNameMap() (map[string]string, error) {
```
jkjell, line 117: Can you change this to *DigestSet? I still don't have a lot of context on this code but, why the name change to `first`?

cryptoutil/digestset.go
```diff
@@ -190,16 +190,16 @@ func CalculateDigestSetFromFile(path string, hashes []crypto.Hash) (DigestSet, e
 	return CalculateDigestSet(file, hashes)
 }
 
-func (ds DigestSet) MarshalJSON() ([]byte, error) {
-	nameMap, err := ds.ToNameMap()
+func (first DigestSet) MarshalJSON() ([]byte, error) {
```
jkjell, line 193: Can you change this to `*DigestSet`?

jkjell (commented):
Thanks for the PR! Just a few minor things and a question.

cryptoutil/digestset.go
naveensrinivasan: Marshal doesn't require a `pointer` receiver. https://pkg.go.dev/encoding/json@go1.19#RawMessage.MarshalJSON

I will revert the name change.

cryptoutil/digestset.go
naveensrinivasan: 👍

cryptoutil/digestset.go
naveensrinivasan: 👍

jkjell (approved):

PR #62, opened 2023-10-30
Improved the search to be concurrent

naveensrinivasan:
- Add support for querying multiple sources concurrently
- Return the combined results from all sources
- If any errors occur, return the first error and discard the results

[source/multi.go]
- Add `sync.WaitGroup` for waiting on all source queries to finish
- Add `errs` and `resChan` channels for collecting errors and results from sources
- Add goroutines for querying sources, collecting errors, and collecting results
- If any errors occurred, return the first error and discard the results
- Return the combined results from all sources


ChaosInTheCRD:
This looks fine on first glance, but given there are no tests it would probably make sense for us to manually test this before merging and approving.

ChaosInTheCRD (approved):

naveensrinivasan:
> After testing this manually it looks like it is working, so I am happy to merge this 😄

Thanks! My apologies have been busy with other things.

PR #63, opened 2023-10-30
Optimized the Search Function

neilnaveen:
- Refactored the `MemorySource` struct in `memory.go` to include `referencesBySubjectDigest`
- Updated the `NewMemorySource` function to initialize these new fields.
- Modified the `LoadEnvelope` function to populate these new fields when loading an envelope.
- Updated the `Search` function to use these new fields when searching for matching envelopes.
- These changes improve the efficiency of the search operation by reducing the time complexity.


ChaosInTheCRD:
@neilnaveen thank you once again for another PR 😄.

I can see that you have certainly made the code more efficient, my only concern is on the side of the complexity of understanding how the maps are operating here with your changes implemented.

I probably will need to pull this down and step through it before accepting. @mikhailswift any thoughts on the changes?

PR #66, opened 2023-11-15
fix: Fix missing checks on product include/exclude glob for attestation.

### LINKED ISSUE #65: Product include and exclude glob is not working correctly.
After some investigation I found that the Attest part of the Product attestor does not exclude or include items specifically. It is taking into account the items when building the list of subjects that is used for creating a subject collection. But it does not check these products during the creation of the attestation.

I test bash script creating different outputs and using the exclude / include glob arguments. But running the code thru the debugger its not using these values during the creation of the Product attestation.

matglas:
This change addresses a problem with `--attestor-product-exclude-glob` and `--attestor-product-include-glob` they only worked for the Subject collection, but not for the Product attestation. Also it did not allow for include overwriting the exclude on sub elements.

Changes:
- It add the two glob arguments to the `RecordArtifacts`. These allow you to pass glob statements the same way they are created in Product attestor for the arguments.
- All references are updated to pass `nil` where the do not have a implementation. This is a breaking change for any API contract for those using this as a library. So maybe I need to change the commit to `fix!:` for breaking change depending on the way we want to maintain contracts.
- The `Product.Subjects()` and `File.shouldRecord()` command now allow for include to take precedence over the exclude.

Fixes #65

### DIFF
--- a/attestation/material/material.go
+++ b/attestation/material/material.go
@@ -69,7 +69,7 @@ func New(opts ...Option) *Attestor {
 }
 
 func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
-	materials, err := file.RecordArtifacts(ctx.WorkingDir(), nil, ctx.Hashes(), map[string]struct{}{})
+	materials, err := file.RecordArtifacts(ctx.WorkingDir(), nil, ctx.Hashes(), map[string]struct{}{}, nil, nil)
 	if err != nil {
 		return err
 	}
--- a/attestation/product/product.go
+++ b/attestation/product/product.go
@@ -164,7 +164,7 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
 	a.compiledExcludeGlob = compiledExcludeGlob
 
 	a.baseArtifacts = ctx.Materials()
-	products, err := file.RecordArtifacts(ctx.WorkingDir(), a.baseArtifacts, ctx.Hashes(), map[string]struct{}{})
+	products, err := file.RecordArtifacts(ctx.WorkingDir(), a.baseArtifacts, ctx.Hashes(), map[string]struct{}{}, compiledIncludeGlob, compiledExcludeGlob)
 	if err != nil {
 		return err
 	}
@@ -194,15 +194,18 @@ func (a *Attestor) Products() map[string]attestation.Product {
 func (a *Attestor) Subjects() map[string]cryptoutil.DigestSet {
 	subjects := make(map[string]cryptoutil.DigestSet)
 	for productName, product := range a.products {
+
+		includeSubject := true
 		if a.compiledExcludeGlob != nil && a.compiledExcludeGlob.Match(productName) {
-			continue
+			includeSubject = false
+		}
+		if a.compiledIncludeGlob != nil && a.compiledIncludeGlob.Match(productName) {
+			includeSubject = true
 		}
 
-		if a.compiledIncludeGlob != nil && !a.compiledIncludeGlob.Match(productName) {
-			continue
+		if includeSubject {
+			subjects[fmt.Sprintf("file:%v", productName)] = product.Digest
 		}
-
-		subjects[fmt.Sprintf("file:%v", productName)] = product.Digest
 	}
 
 	return subjects
--- a/attestation/file/file.go
+++ b/attestation/file/file.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 
 	"github.com/edwarnicke/gitoid"
+	"github.com/gobwas/glob"
 	"github.com/testifysec/go-witness/cryptoutil"
 	"github.com/testifysec/go-witness/log"
 )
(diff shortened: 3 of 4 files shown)


mikhailswift:
Hi @matglas, thanks for the contribution!

I should explain my thought process behind this decision to determine the best course of action. Long story short, I intended to still include files in the product and material attestors so we could still use them for integrity checking between steps. The thought is that it would be less than desirable if I could fool a policy to sneak files into a compilation process by just including a CLI argument to ignore those files.

However, I can still imagine use cases where we may want to exclude files from the materials and product attestors. And perhaps an implementation of in-toto's [artifact rules](https://github.com/in-toto/docs/blob/v1.0/in-toto-spec.md#433-artifact-rules) would help mitigate my concerns.

I know from experience that it's painful to create an attestation about an `npm install` and get a huge attestation output due to the sheer number of files included in the product attestor.

matglas:
In my case we are building a mono repo. So thats already a lot of file to index and checksum. Then the output of our build tool, https://please.build, has intermediate output of different targets which is there inside a plz-out folder but it also has all the intermediate stuff that is needed to build the final artifacts. That is not one project with node modules but multiple. If I don't use the glob like I implemented my output file is over a 1gb in size and take almost ~10 minutes to create.

matglas:
To add the final thing that I want to verify is the a package of all the things so I would not need the intermediate files. They are in the package. But with the current setup I have for executing build in the pipeline I do not have the option yet to remove the intermediates after I packaged it.

I might be able to add a wrapper on the build command and only keep what I need.

matglas:
I did notice my logic in the code is not fully working. If you knly specify the exclude it still includes everything because the include is a `*` by default.

matglas:
I implemented the logic to fix the default glob combined with an exclude. And I added configuration output so a policy can reason about if it wants to accept this configuration. Because of the output change in the attestation I bumped the version number too.

matglas:
Hi @jkjell I see you pushed changes. But because I'm not too familiar with github and this process yet I can't really understand what happened. Can you explain what you did?

I just pushed some more changes. And merged main back to my branch if I'm correct.

matglas:
Also I noticed that the test jobs for product attestor fail. I need to fix those.

matglas:
Did an rebase to sign-off all commits.

attestation/product/product.go
```diff
@@ -103,6 +103,16 @@ type Attestor struct {
 	compiledExcludeGlob glob.Glob
 }
 
+type attestorJson struct {
+	Products      map[string]attestation.Product `json:"products"`
+	Configuration attestorConfiguration          `json:"configuration"`
```
jkjell, line 108:
```suggestion
	Configuration *attestorConfiguration          `json:"configuration,omitempty"`
```

I think it would be nice to exclude the config if no globs are set.

attestation/product/product.go
```diff
@@ -174,16 +184,28 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
 }
 
 func (a *Attestor) MarshalJSON() ([]byte, error) {
-	return json.Marshal(a.products)
+	output := attestorJson{
+		Products: a.products,
+		Configuration: attestorConfiguration{
```
jkjell, line 189: If we make Configuration optional, we can construct the configuration prior to this and check if both globs are `""`.

attestation/material/material.go
```diff
@@ -36,15 +42,68 @@ var (
...
+	materials           map[string]cryptoutil.DigestSet
+	includeGlob         string
+	compiledIncludeGlob glob.Glob
+	excludeGlob         string
+	compiledExcludeGlob glob.Glob
+}
+
+type attestorJson struct {
+	Materials     map[string]cryptoutil.DigestSet `json:"materials"`
+	Configuration attestorConfiguration           `json:"configuration"`
```
jkjell, line 101: Same comment here as the product attestor.

jkjell (commented):
I think this is just about ready! Just a minor suggestion to allow the Configuration to be omitted if no globs are used.

attestation/product/product.go
matglas: Yes that possible. I'll check on that.

attestation/material/material.go
```diff
@@ -16,16 +16,22 @@ package material
...
 	"github.com/in-toto/go-witness/attestation"
 	"github.com/in-toto/go-witness/attestation/file"
 	"github.com/in-toto/go-witness/cryptoutil"
+	"github.com/in-toto/go-witness/registry"
 )
 
 const (
 	Name    = "material"
-	Type    = "https://witness.dev/attestations/material/v0.1"
+	Type    = "https://witness.dev/attestations/material/v0.2"
```
matglas, line 30: @jkjell I was just thinking if we might need to mention somewhere that the version is bumped. Like a changelog of some sorts? This will break current policies that are in place. Same goes for the product.

attestation/product/product.go
```diff
@@ -32,7 +32,7 @@ import (
 
 const (
 	Name    = "product"
-	Type    = "https://witness.dev/attestations/product/v0.1"
+	Type    = "https://witness.dev/attestations/product/v0.2"
```
matglas, line 35: Same as material attestor comment about changelog.

matglas: I added CHANGELOG-ATTESTORS.md.

attestation/material/material.go
matglas: I added CHANGELOG-ATTESTORS.md.

attestation/material/material.go
matglas: Fixed this

attestation/product/product.go
matglas: Fixed this.

attestation/material/material.go
```diff
@@ -100,16 +171,36 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
...
+		}
+		if a.excludeGlob != "" {
+			config.ExcludeGlob = a.excludeGlob
+		}
+	}
+
+	return json.Marshal(output)
 }
 
 func (a *Attestor) UnmarshalJSON(data []byte) error {
```
jkjell, line 192: We'll need to make this backwards compatible. I think we can check `a.Type()` for the version and unmarshal into the new struct format.

jkjell:
@matglas I brought everything up-to-date and resolved the conflicts. I added one more comment about backwards compatibility. I think if we can get that and testing to ensure it works, we should be good. On resolving the conflicts, there was some overlap with a tracing PR that @mikhailswift reviewed. He's back tomorrow so, it'll be good to have him take a look and make sure I didn't break that either. ☺️

colek42:
@mikhailswift did you get a chance to look at this?

mikhailswift:
I tried to reconcile all the conflicts. Can you take a look at #656 to see if it looks like everything got brought in right? @matglas

PR #83, opened 2023-11-29
 Refactoring error messages to use %w formatting directive and fix logging issue

ChaosInTheCRD:
This PR coincides with PR https://github.com/in-toto/witness/pull/314 and aims to standardised the error formatting across the witness codebase.

I have tried my best to make sure these changes are standardised across the entire repository, so if anyone would like to review and double check that would be much appreciated 😄.

### DIFF
--- a/attestation/aws-iid/aws-iid.go
+++ b/attestation/aws-iid/aws-iid.go
@@ -134,12 +134,12 @@ func (a *Attestor) getIID() error {
 	svc := ec2metadata.New(&a.session, a.conf)
 	iid, err := svc.GetDynamicData(docPath)
 	if err != nil {
-		return fmt.Errorf("failed to get instance identity document: %v", err)
+		return fmt.Errorf("failed to get instance identity document: %w", err)
 	}
 
 	sig, err := svc.GetDynamicData(sigPath)
 	if err != nil {
-		return fmt.Errorf("failed to get signature: %v", err)
+		return fmt.Errorf("failed to get signature: %w", err)
 	}
 
 	a.RawIID = iid
@@ -147,7 +147,7 @@ func (a *Attestor) getIID() error {
 
 	err = json.Unmarshal([]byte(a.RawIID), &a.EC2InstanceIdentityDocument)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal iid: %v", err)
+		return fmt.Errorf("failed to unmarshal iid: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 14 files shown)


attestation/aws-iid/aws-iid.go
```diff
@@ -182,12 +182,13 @@ func (a *Attestor) Verify() error {
...
 	if err != nil {
-		return fmt.Errorf("failed to encode public key: %v", err)
+		return fmt.Errorf("failed to encode public key: %w", err)
 	}
 
 	err = rsa.VerifyPKCS1v15(pubKey, crypto.SHA256, docHash[:], sigBytes)
 	if err != nil {
-		log.Debugf("(attestation/aws-iid) failed to verify signature: %v", err)
+		err := fmt.Errorf("(attestation/aws-iid) failed to verify signature: %w", err)
+		log.Debug(err)
```
jkjell, line 191: Would this work?

```log.Errorf("(attestation/aws-iid) failed to verify signature: %w", err)```

jkjell (commented):
If the one proposed change doesn't work, LGTM. 👍

attestation/aws-iid/aws-iid.go
ChaosInTheCRD: So I've just done a bit more digging on this and it seems I slightly misconveyed things.

`go-witness` has [package log](https://github.com/in-toto/go-witness/blob/main/log/log.go) which is an interface to allow different logging libraries to be plugged in. 

We use `logrus` in https://github.com/in-toto/witness and _I believe_ `log.Errorf` in itself does not support error wrapping with the use of `%w`. This is why I decided to use `fmt.Errorf` separately. Instead, it allows for the use of `WithError` and from what I can tell, expects you to wrap errors that way.

We could work round this by expanding the package so the interface implements `WithError`, but that would leave those wanting to leverage the logging inside the `go-witness` library needing to implement this themselves. This (I think) also holds true if we wish to more deeply use more of the `logrus` library's features along the line.

The way I see it (and @jkjell would love to know your opinion along with others), we have a few options:
a) expand the interface whenever we need so we can use logrus features. This seems like a bad idea to me as if those utilising the library are using other loggers, it would be a pain.
b) handle this (and other situations like it) with some utility function inside the `log` package. This actually seems the right way to me, I can neaten all this up right now by adding this to the package. This could possibly lead to situations where we are handing things back to the logger in an incorrect form though? Not sure if that is a problem.
c) Don't do any logging inside the library. I think this would be troublesome for Debug logs, I know for a fact that logging implemented inside `cosign` libraries has come in handy in the past when I have been doing projects.

Any thoughts welcome 😄

jkjell: @mikhailswift what are your opinions here?

mikhailswift: I am heavily leaning toward b, and I think @ChaosInTheCRD 's reasoning is sound. 

The reasoning for implementing the log interface within the go-witness library was to keep it generic enough for users of the library to use their logging library of choice -- I didn't want the library to be too opinionated. Libraries logging to stdout or stderr unexpectedly creates a frustrating experience for library users. c kind of hints at this, there's a balance where you don't want a library to log on you, but debug logging can be extraordinarily useful.

kriscoleman: great stuff @ChaosInTheCRD! 

I also like the sounds of B, try to reduce the amount of pain our users feel

ChaosInTheCRD: So I figured what we could do is check for an error in the relevant `f` functions (e.g., `Debugf`, `Warnf`) for arguments of type `error`, and handle that inside the wrapper function. I have it in a new branch (to get rid of the now unnecessary refactoring) [here](https://github.com/ChaosInTheCRD/go-witness/blob/refactoring-error-wrapping/log/log.go#L69-L76).

Thoughts would be appreciated. If this works I can go ahead and close this PR and I'll raise a PR for the other branch 😄 .

ChaosInTheCRD:
Closing this PR in favour of https://github.com/in-toto/go-witness/pull/83

PR #97, opened 2023-12-11
Updating README

ChaosInTheCRD:
Since this separate library has existed, the README has been a copy of the main repository README.

While there is a lot of work to do on refactoring documentation, I think simplifying this README is a nice start. This makes it easier to understand what the repository serves as, and also removes the current problems observed with references to documentation in this README.

At a certain point we will aim to consolidate all witness related documentation into a single docs source location and expose that with a public frontend.


README.md
```diff
@@ -1,302 +1,32 @@
...
-- Notary v2 Signature Validation
-- [Zarf](https://github.com/defenseunicorns/zarf) Integration
-- IronBank Attestor
-
-## Support
-
-[TestifySec](https://testifysec.com) Provides support for witness and other CI security tools.
-[Contact Us](mailto:info@testifysec.com)
+## Status
+This library is currently pre-1.0 and therefore should be considered unstable and subject to change.
```
jkjell, line 11: > This library is currently pre-1.0 and therefore should be considered unstable and subject to change.

Maybe to something like: 
> This library is currently pre-1.0 and therefore the API may be subject to breaking changes.

README.md
```diff
@@ -1,302 +1,32 @@
...
+
+## Documentation
+For more detail regarding the library itself, we recommend viewing [pkg.go.dev](https://pkg.go.dev/github.com/testifysec/go-witness). For
+the documentation of the witness project, please view [the main witness repository](https://github.com/in-toto/witness/tree/main/docs).
+
+## Requirements
+In order to effectively contribute to this library, you will need:
+- A Unix-compatible Operating System
+- GNU Make
+- Go 1.19
```
jkjell, line 29: Side note: we should chat about minimum support Go version at some point. I think we use Go 1.21.x in most places.

jkjell (dismissed):

README.md
ChaosInTheCRD: yeah for sure - I was actually looking to tackle that when getting the development environment configuration figured out and defined.

README.md
```diff
@@ -1,302 +1,32 @@
-[![OpenSSF Best Practices](https://www.bestpractices.dev/projects/8164/badge)](https://www.bestpractices.dev/projects/8164)
+# go-witness
+A client library for [Witness](https://github.com/in-toto/witness), written in Go.
+
+[![Go Reference](https://pkg.go.dev/badge/github.com/in-toto/go-witness.svg)](https://pkg.go.dev/github.com/in-toto/go-witness)
+[![Go Report Card](https://goreportcard.com/badge/github.com/in-toto/go-witness)](https://goreportcard.com/report/github.com/in-toto/go-witness)
+![OpenSSF Best Practices](https://www.bestpractices.dev/projects/8164/badge)](https://www.bestpractices.dev/projects/8164)
```
jkjell, line 6: Miss `[` at the beginning of the line.

jkjell (changes requested):

README.md
ChaosInTheCRD: not sure how I managed that

jkjell (approved):

PR #104, opened 2023-12-15
Adding function for adding a single attestor

ChaosInTheCRD:
This change has been added so we can add one attestor at a time. See https://github.com/in-toto/witness/pull/332 for use case.


attestation/factory.go
```diff
@@ -86,7 +86,15 @@ func FactoryByName(name string) (registry.FactoryFunc[Attestor], bool) {
...
-func Attestors(nameOrTypes []string) ([]Attestor, error) {
+func AddAttestor(nameOrType string) (Attestor, error) {
+	attestors, err := AddAttestors([]string{nameOrType})
+	if err != nil {
+		return nil, err
+	}
+	return attestors[0], nil
+}
+
+func AddAttestors(nameOrTypes []string) ([]Attestor, error) {
```
kairoaraujo, line 97: Overall, the changes are pretty simple; I have a question about library API compatibility.

I see it breaks compatibility with previous versions. (I see you fix it in https://github.com/in-toto/witness/pull/332)
How will we document for other users of go-witness?

kairoaraujo (commented):
Overall, the changes are pretty simple; I have a question about library API compatibility.

attestation/factory.go
jkjell: Yeah, I think we would need to keep a `func Attestors` with a deprecation notice that just calls AddAttestors.

ChaosInTheCRD:
Just gone and made a mess of this by signing off a load of commits that aren't mine 🤦 going to close this PR and reopen on a fresh branch as it's only a tiny change

attestation/factory.go
mikhailswift: One quick thing -- The `Attestors` function currently would be more equivalent to a `GetAttestors` function -- it's returning a slice of instantiated Attestors from the factory.

`AddAttestor`, to me, seems to imply we're adding more attestors to the factory.

PR #108, opened 2023-12-17
Update SECURITY-INSIGHTS.yml with additional information

jkjell:
- [x]  Dependency Policy
- [ ]  SBOM
- [x]  Security Insights file
- [x]  Token Permissions


jkjell:
Will address SBOM as a part of a separate change to the release process.

PR #109, opened 2023-12-18
WIP: Defining an interface for allowing signing and verifying with different envelope types

ChaosInTheCRD:
This is a draft for now so people can see my progress.

I am trying to refactor the `dsse` and `cryptoutil` parts of the repo so they can be:
a) set up for the expansion to more envelope types to sign and verify
b) make the current internal `dsse` library more streamlined with https://github.com/secure-systems-lab/go-securesystemslib. While we don't use it right now, we plan on migrating to it when it supports timestamps and intermediates (I think this PR might be related https://github.com/secure-systems-lab/dsse/pull/61).

### DIFF
--- a/cryptoutil/ecdsa.go
+++ b/cryptoutil/ecdsa.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rand"
+	"crypto/x509"
 )
 
 type ErrVerifyFailed struct{}
@@ -38,6 +39,14 @@ func NewECDSASigner(priv *ecdsa.PrivateKey, hash crypto.Hash) *ECDSASigner {
 
 func (s *ECDSASigner) KeyID() (string, error) {
 	return GeneratePublicKeyID(&s.priv.PublicKey, s.hash)
+}
+
+func (s *ECDSASigner) Algorithm() (x509.PublicKeyAlgorithm, crypto.Hash) {
+	return x509.ECDSA, s.hash
+}
+
+func (s *ECDSASigner) Signer() (crypto.Signer, error) {
+	return s.priv, nil
 }
 
 func (s *ECDSASigner) Sign(ctx context.Context, data []byte) ([]byte, error) {
--- a/cryptoutil/ecdsa.go
+++ b/cryptoutil/ecdsa.go
@@ -15,10 +15,10 @@
 package cryptoutil
 
 import (
+	"context"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rand"
-	"io"
 )
 
 type ErrVerifyFailed struct{}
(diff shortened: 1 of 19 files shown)


cryptoutil/ecdsa.go
```diff
@@ -40,8 +40,8 @@ func (s *ECDSASigner) KeyID() (string, error) {
 	return GeneratePublicKeyID(&s.priv.PublicKey, s.hash)
 }
 
-func (s *ECDSASigner) Sign(r io.Reader) ([]byte, error) {
-	digest, err := Digest(r, s.hash)
+func (s *ECDSASigner) Sign(ctx context.Context, data []byte) ([]byte, error) {
```
colek42, line 43: does this need to be depricated?

cryptoutil/rsa.go
```diff
@@ -65,8 +65,8 @@ func (v *RSAVerifier) KeyID() (string, error) {
 	return GeneratePublicKeyID(v.pub, v.hash)
 }
 
-func (v *RSAVerifier) Verify(data io.Reader, sig []byte) error {
-	digest, err := Digest(data, v.hash)
+func (v *RSAVerifier) Verify(ctx context.Context, data []byte, sig []byte) error {
```
colek42, line 68: how will using []byte vs io.Reader effect memory usage?  Verify is likely to be used by a serivce

cryptoutil/signer.go
```diff
@@ -32,16 +33,23 @@ func (e ErrUnsupportedKeyType) Error() string {
...
-	Sign(r io.Reader) ([]byte, error)
-	Verifier() (Verifier, error)
+type SignerVerifier struct {
+	Signer
+	Verifier
+}
+
+// NOTE: Signer verifier required this function and I don't know why
+func (sv SignerVerifier) KeyID() (string, error) {
+	return sv.Signer.KeyID()
```
colek42, line 43: the keyID needs to be Canonical

cryptoutil/ecdsa.go
ChaosInTheCRD: I presume you mean the switch from taking in an `io.Reader` vs `Context` and `[]byte`?

So this hasn't been documented at all yet, and probably needs doing so, but switching to this is to align closer to the [secure systems lab lib](https://github.com/secure-systems-lab/dsse), so moving over to it is less hassle when we do.

However, passing in an `io.Reader` might be more beneficial for making all of this test driven, so it probably warrants a wider discussion. Also there might be other perfectly good reasons for this being passed as an argument.

To answer the original question, I am not sure about the need for deprecation inside this `cryptoutil` library more generally. Are we intending for it to be used directly? I have also been thinking about embedding it within the `signature` package, and only exposing it in the form of `NewSignerFromReader`/`NewSignerFromFile`. Not sure what your thoughts are on this?

cryptoutil/rsa.go
ChaosInTheCRD: good question - one that I haven't thought of. The main motivation was to streamline it with the secure systems lab dsse implementation.

signature/envelope/cose/cose.go
```diff
@@ -0,0 +1,108 @@
+package cose
```
colek42, line 1: COSE has a draft spec for RFC 3161 ref:  https://www.ietf.org/id/draft-ietf-cose-tsa-tst-header-parameter-01.html

We should get this implemented as part of the first release

ChaosInTheCRD:
Given the aims to consolidate this library against other go libraries within in-toto and secure-systems-lib (https://github.com/secure-systems-lab/go-securesystemslib and https://github.com/in-toto/in-toto-golang), we are going to hold off on finishing and merging this work. It will however come in handy when the time comes.

PR #120, opened 2024-01-08
KMS Support

ChaosInTheCRD:
### In this PR
- [x] AWS KMS Signer / Verifier
- [x] AWS KMS Signer / Verifer Tests
- [x] GCP KMS Signer / Verifier
- [x] GCP KMS Signer / Verifier Tests

### In a follow-up PR
- [ ] Azure KMS Signer / Verifier
- [ ] Vault KMS Signer / Verifier


adityasaky:
In light of go-consolidation, would it make sense to explore putting it in go-sslib directly?

jkjell:
> would it make sense to explore putting it in go-sslib directly

This is a good question. We should have a plan for incremental adoption of go-ssllib. We'll think about this before next Friday's call.

jkjell:
We have some demand for this from users that are interested before the go-ssl work is likely to be done.

colek42:
> In light of go-consolidation, would it make sense to explore putting it in go-sslib directly?

@adityasaky, I think that is a path we would love to explore.  However, I the project needs to be under open governance.  Let's discuss at a community meeting or KubeCon EU with the rest of the team.

colek42:
Would it make sense to breat these into 2 commits one for each signer verifier?

adityasaky:
> @adityasaky, I think that is a path we would love to explore. However, I the project needs to be under open governance. Let's discuss at a community meeting or KubeCon EU with the rest of the team.

We've (cc @jkjell) started working towards this. The idea is a lightweight go-sslib governance that gives maintainer bits to TUF, in-toto etc.

registry/registry.go
```diff
@@ -81,7 +81,16 @@ func (r Registry[T]) AllEntries() []Entry[T] {
 	}
 
 	return results
+}
 
+// AllEntries returns every Entry in the Registry that can be used as a Verifier
```
mikhailswift, line 86: The registry isn't Verifier specific, it's just a generic factory pattern. I don't think you need this function, you should be able to use the existing `AllEntries` function

signer/kms/aws/client.go
```diff
@@ -0,0 +1,341 @@
...
+	case types.SigningAlgorithmSpecRsassaPssSha384, types.SigningAlgorithmSpecRsassaPkcs1V15Sha384, types.SigningAlgorithmSpecEcdsaSha384:
+		return crypto.SHA384
+	case types.SigningAlgorithmSpecRsassaPssSha512, types.SigningAlgorithmSpecRsassaPkcs1V15Sha512, types.SigningAlgorithmSpecEcdsaSha512:
+		return crypto.SHA512
+	default:
+		return 0
+	}
+}
+
+func (c *cmk) Verifier() (cryptoutil.Verifier, error) {
```
mikhailswift, line 318: I think most of this function can just be replaced with a call to [cryptoutil.NewVerifier](https://github.com/in-toto/go-witness/blob/1f9f95edcb8108cdcf8a1f94d723688e88919126/cryptoutil/verifier.go#L67).

This switch case seemingly boils down to a type assertion of c.PublicKey and instantiating the appropriate cryptoutil.Verifier depending on the results -- which is what NewVerifier does.

So I think you can just do `return cryptoutil.NewVerifier(c.PublicKey, cryptoutil.VerifyWithHash(c.HashFunc())`

signer/kms/aws/go.mod
```diff
@@ -0,0 +1,33 @@
+module github.com/in-toto/go-witness/signer/kms/aws
```
mikhailswift, line 1: I don't believe we want a nested go module here

signer/kms/aws/signer.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+
+	return cryptoutil.PublicPemBytes(p)
+}
+
+// Verify verifies the signature for the given message, returning
+// nil if the verification succeeded, and an error message otherwise.
+func (a *SignerVerifier) Verify(message io.Reader, sig []byte) (err error) {
+	ctx := context.Background()
```
mikhailswift, line 128: I think long term we'll want to change the interface to allow for passing the context in directly. Contexts can be tied to specific requests and used to cancel requests to external resources -- which we want here.

We can punt on it for now, but I'd recommend changing these `context.Background` calls to `context.TODO()` instead for the time being

mikhailswift:
I'll continue my review tomorrow as well as testing it all locally. I'm curious about the nested go.mods. These have the capability to cause a _ton_ of pain, and I think we should avoid it in favor of just having a singular top level go.mod.

I have some other thoughts but will roll them into further review comments.

Very good work so far!

signer/kms/gcp/client.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	var err error
+	g.projectID, g.locationID, g.keyRing, g.keyName, g.version, err = parseReference(ksp.Reference)
+	if err != nil {
+		return nil, err
+	}
+
+	// NOTE: We need to think about how we want to handle passing in options here
+	g.client, err = gcpkms.NewKeyManagementClient(ctx)
```
mikhailswift, line 128: We can use the functional option pattern we use elsewhere to supply KeyManagementClientOptions

```
func WithKeyManagementClientOptions(opts ...gcpkms.ClientOption) Option {
    return func() { ... }
}
```

We can pair up on this if you like :)

signer/kms/gcp/client.go
```diff
@@ -0,0 +1,399 @@
...
+	// kv is keyVersion to use
+	crv := cryptoKeyVersion{
+		CryptoKeyVersion: kv,
+		PublicKey:        pubKey,
+	}
+
+	// crv.Verifier is set here to enable storing the public key & hash algorithm together,
+	// as well as using the in memory Verifier to perform the verify operations.
+	switch kv.Algorithm {
+	case kmspb.CryptoKeyVersion_EC_SIGN_P256_SHA256:
```
mikhailswift, line 215: similar to my comment on the AWS verifier, I think some of this can be simplified to just picking up which hash to use and calling `cryptoutil.NewVerifier` with the `VerifyWithHash` option

signer/kms/aws/signer.go
```diff
@@ -0,0 +1,165 @@
...
+
+	var signerOpts crypto.SignerOpts
+	signerOpts, err = a.client.getHashFunc(ctx)
+	if err != nil {
+		return fmt.Errorf("getting hash func: %w", err)
+	}
+	hf := signerOpts.HashFunc()
+
+	// we want to verify remotely unless we explicitly ask to verify locally
+	if os.Getenv("WITNESS_AWS_VERIFY_LOCALLY") == "1" {
```
mikhailswift, line 139: Generally I don't think we should be reaching into the env directly here -- instead I think this option should be supplied to the `SignerVerifier` struct at creation with the Option pattern, and from there we can add support for configuring these via the environment.  I have some thoughts on this we should sync on.

signer/kms/aws/client.go
```diff
@@ -0,0 +1,341 @@
...
+
+	item := a.keyCache.Get(cacheKey, ttlcache.WithLoader[string, cmk](loader))
+	if lerr == nil {
+		cmk := item.Value()
+		return &cmk, nil
+	}
+	return nil, lerr
+}
+
+// At the moment this function lies unused, but it is here for future if necessary
```
kriscoleman, line 215: food for thought: my yagni practices tell me maybe this shouldn't be here if it's unused

signer/kms/aws/signer.go
kriscoleman: I have sometimes been burned by env vars that aren't discoverable in documentation but you glean from source code. This seems like a good pattern for intuitive dev ex

signer/kms/aws/signer_test.go
```diff
@@ -0,0 +1,374 @@
...
+				t.Fatal("expected verification to fail")
+			}
+
+		}
+
+		if tt.wantErr && !errFound {
+			t.Fatalf("expected test %s to fail with error %s", tt.name, tt.expectedErr.Error())
+		}
+	}
+}
```
kriscoleman, line 374: I really love the test patterns you used here

signer/kms/aws/client.go
ChaosInTheCRD: yeah the only reason that I kept this the way that it is is the fact that this is checking against the AWS types directly. Maybe this is simply unnecessary, provided that PKCS #1 v1.5 keys work as they should? I'll check it out

ChaosInTheCRD: yeah okay looks like we're all good here. Just ran and everything *seems* happy. Although not sure if we will hit some unexpected edge cases.

ChaosInTheCRD: Do you think it would be worth adding some steps with different signer / verifier types @mikhailswift ?

signer/kms/aws/go.mod
ChaosInTheCRD: so we can remove the nested go module here. However it's probably worth us referring back to the sigstore folks to ask why the nested go module was implemented here. Maybe they have a reason why they did this 😄. I can't register the kms providers in `kms/signerprovider.go` because it leads to an import cycle, so still need to reference them in `cmd/root.go` of https://github.com/in-toto/witness`.

signer/kms/aws/signer.go
ChaosInTheCRD: Yeah I'lll set this as a `TODO()` for myself!

signer/kms/gcp/client.go
ChaosInTheCRD: yeah so my NOTE here is more along the lines of making sure how we consider the UX of presenting the options in the CLI. I am just wanting to avoid the `--everything` dilemma. Do you think that environment variables would be an acceptable way of exposing options?

signer/kms/gcp/client.go
ChaosInTheCRD: yeah okay sure, I've modified and tested for AWS, just need to do the same for GCP.

signer/kms/aws/signer.go
ChaosInTheCRD: sure thing. I (like with other comments) am thinking about reducing the flag toil, but in reality i'm in agreement that these env vars can lead to burns.

signer/kms/aws/client.go
ChaosInTheCRD: yeah agreed.

signer/kms/aws/signer_test.go
ChaosInTheCRD: huh - I had imposter syndrome that this was terrible! This comment made me feel good <3

ChaosInTheCRD:
Still TODO befoire merge:

- [ ] Write tests to ensure that different key types from GCP/AWS are being parsed correctly by `cryptoutil`
- [ ] Implement options for AWS and GCP KMS Signers
- [ ] Add placeholder contexts in Signer functions

signer/kms/aws/client.go
ChaosInTheCRD: ah - this comment is now incorrect. It is used for offline verification 😄

signer/kms/aws/signer.go
mikhailswift: We do need to support configuration through env, but I think we have an opportunity to make it more generic across the board so we don't have to have hard coded flags strewn about

signer/kms/aws/client.go
```diff
@@ -0,0 +1,318 @@
...
+		return crypto.SHA384
+	case types.SigningAlgorithmSpecRsassaPssSha512, types.SigningAlgorithmSpecRsassaPkcs1V15Sha512, types.SigningAlgorithmSpecEcdsaSha512:
+		return crypto.SHA512
+	default:
+		return 0
+	}
+}
+
+func (c *cmk) Verifier() (cryptoutil.Verifier, error) {
+	return c.Verifier()
```
jkjell, line 317: Looks like this entire function was taken out. All that's left is a recursive call. 😊

signer/kms/aws/signer.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return a, nil
+}
+
+// NOTE: This might be all wrong but setting it like so for now
+//
+// KeyID returns the key identifier for the key used by this signer.
+func (a *SignerVerifier) KeyID() (string, error) {
+	return a.reference, nil
+}
```
jkjell, line 83: Is this the right KeyID to use? From the awsClient in client.go: https://github.com/in-toto/go-witness/pull/120/files#diff-5679de7065f19bd20365541d52c8835494a381507a05877b4f34569d62a7748fR115

jkjell: Additionally, should we have a `PublicKey()` function too?

jkjell (commented):
I didn't dive in too much to gcp or hashivault.

ChaosInTheCRD:
> I didn't dive in too much to gcp or hashivault.

dang, sorry if I messed anyone around 🤦 - hashicorp vault KMS is being held back for a further release. I removed it from this PR, sorry for the inconvenience.

signer/kms/aws/signer.go
ChaosInTheCRD: aha - sorry @jkjell this has gotten lost in other comments and actually poses a question:
> Should the KeyID be the KMS reference?

At the moment a lot of the functionality (e.g., defining KMS in the policy) relies on the KeyID being the KMS reference.

jkjell (dismissed):
Great work!

cryptoutil/util.go
```diff
@@ -85,6 +96,23 @@ func PublicPemBytes(pub interface{}) ([]byte, error) {
...
+func UnmarshalPEMToPublicKey(pemBytes []byte) (crypto.PublicKey, error) {
+	derBytes, _ := pem.Decode(pemBytes)
+	if derBytes == nil {
+		return nil, errors.New("PEM decoding failed")
+	}
+	switch derBytes.Type {
+	case string(PublicKeyPEMType):
+		return x509.ParsePKIXPublicKey(derBytes.Bytes)
+	case string(PKCS1PublicKeyPEMType):
+		return x509.ParsePKCS1PublicKey(derBytes.Bytes)
```
mikhailswift, line 109: Can we add a call to `ParsePKCS1PublicKey` in the `TryParsePEMBlock` function?

mikhailswift (dismissed):
great work! we can handle my comment in a separate PR

mikhailswift (approved):

PR #124, opened 2024-01-12
Adding support for using timestamp authority and CA certificates for verifying policy

ChaosInTheCRD:
After discovering that the `policy-ca` flag in `github.com/in-toto/witness` was not wired up correctly, I noticed that some change was required in go-witness to feed it into the verification function for the policy DSSE envelope. I also noticed that there was no way of specifying a timestamp authority CA certificate for the policy. Given that `Sign` supports specifying said timestamp authority, it seemed clear that adding support in Verify would be worthwhile.

Note that currently, the way that we are verifying the policy is different from how we verify attestations. It seems that said disparity does not need to be there and I will therefore create another Issue to track the need to refactor this. 

See Github issue here: https://github.com/in-toto/go-witness/issues/125

### DIFF
--- a/dsse/dsse.go
+++ b/dsse/dsse.go
@@ -32,11 +32,11 @@ func (e ErrNoMatchingSigs) Error() string {
 
 type ErrThresholdNotMet struct {
 	Theshold int
-	Acutal   int
+	Actual   int
 }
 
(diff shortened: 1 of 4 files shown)


jkjell (approved):
Looks good!

PR #128, opened 2024-01-16
Adding function to add a single attestor

ChaosInTheCRD:
allows a single attestor to be added rather than multiple as input to function. See https://github.com/in-toto/go-witness/pull/104 for more details.

### DIFF
--- a/attestation/factory.go
+++ b/attestation/factory.go
@@ -86,7 +86,20 @@ func FactoryByName(name string) (registry.FactoryFunc[Attestor], bool) {
 	return registrationEntry.Factory, ok
 }
 
+func AddAttestor(nameOrType string) (Attestor, error) {
+	attestors, err := AddAttestors([]string{nameOrType})
+	if err != nil {
+		return nil, err
+	}
+	return attestors[0], nil
+}
+
+// Deprecated: use AddAttestors instead
 func Attestors(nameOrTypes []string) ([]Attestor, error) {
+	return AddAttestors(nameOrTypes)
+}
+
+func AddAttestors(nameOrTypes []string) ([]Attestor, error) {
 	attestors := make([]Attestor, 0)
 	for _, nameOrType := range nameOrTypes {
 		factory, ok := FactoryByName(nameOrType)
--- a/attestation/factory.go
+++ b/attestation/factory.go
@@ -70,6 +70,12 @@ func (e ErrAttestationNotFound) Error() string {
 	return fmt.Sprintf("attestation not found: %v", string(e))
 }
 
+type ErrAttestorNotFound string
+
+func (e ErrAttestorNotFound) Error() string {
+	return fmt.Sprintf("attestor not found: %v", string(e))
+}
+
 func RegisterAttestation(name, predicateType string, run RunType, factoryFunc registry.FactoryFunc[Attestor], opts ...registry.Configurer) {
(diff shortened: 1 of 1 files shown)


attestation/factory.go
```diff
@@ -86,7 +86,20 @@ func FactoryByName(name string) (registry.FactoryFunc[Attestor], bool) {
 	return registrationEntry.Factory, ok
 }
 
+func AddAttestor(nameOrType string) (Attestor, error) {
```
mikhailswift, line 89: One quick thing -- The Attestors function currently would be more equivalent to a GetAttestors function -- it's returning a slice of instantiated Attestors from the factory.

AddAttestor, to me, seems to imply we're adding more attestors to the factory. Maybe `GetAttestor` or just `Attestor` is an appropriate name? I tend to avoid `GetXXXX` named functions but that's just my 2 cents.

attestation/factory.go
```diff
@@ -86,7 +86,20 @@ func FactoryByName(name string) (registry.FactoryFunc[Attestor], bool) {
...
+	attestors, err := AddAttestors([]string{nameOrType})
+	if err != nil {
+		return nil, err
+	}
+	return attestors[0], nil
```
mikhailswift, line 94: We may want to check that `len(attestors) > 0` prior to returning here to prevent a possible panic

attestation/factory.go
ChaosInTheCRD: Sorry if I am getting this wrong, but I was under the impression that `AddAttestor(s)` made sense here because we are providing a `string` which defines the attestor we wish to instantiate, it instantiates it and *then* returns it. `GetAttestor` suggests (to me at least) that we are fetching something that has already been instantiated, hence why I went with `AddAttestor`.

ChaosInTheCRD: oh scrap that, I see what you mean.

ChaosInTheCRD: I think that while I am impartial with whether at this stage it should be `Attestors` or `AddAttestors`, that isn't enough to go and change the function name. I will revert it back to the original naming.

mikhailswift: I'm not married to the naming, for what it's worth. Just giving ya my thought process :)

ChaosInTheCRD: Ah I see. If the new function is `func Attestor` that collides with the interface. Let me know if you're happy with me keeping it as `AddAttestors` or if you wish for it to be named `GetAttestors`

mikhailswift: I think `GetXXXX` makes more sense to me in this context, but I am curious about what others things @jkjell

attestation/factory.go
ChaosInTheCRD: This has now prompted me to wonder. We currently use https://github.com/in-toto/go-witness/blob/404b65482adde0bdff021c20b6c49eab543d9e56/attestation/factory.go#L69-L71 if we can't find the attestor. Wouldn't this be better as a separate error saying the `Attestor` can't be found, rather than the `attestation`?

attestation/factory.go
mikhailswift: Or, since we use `New` in other contexts to indicate we're creating a new instance of an object, we could use `NewAttestor` and `NewAttestors` .

ChaosInTheCRD: I am happy to switch this to `GetAttestor(s)` 😃

mikhailswift (approved):
good work!

PR #131, opened 2024-01-22
RunAttestors refactor

ChaosInTheCRD:
Small work done to refactor the `RunAttestors()` function.

I have also made some changes to the `AttestationContext` functions, I am not quite sure why copies of the struct fields being returned need to be made. @mikhailswift I'll request your review to try and check that there isn't any extra context on this.

### DIFF
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -125,6 +125,8 @@ func (ctx *AttestationContext) RunAttestors() error {
 				Reason: fmt.Sprintf("unknown run type %v", attestor.RunType()),
 			}
 		}
+
+		attestors[attestor.RunType()] = append(attestors[attestor.RunType()], attestor)
 	}
 
 	for _, atts := range attestors {
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -117,64 +117,23 @@ func NewContext(attestors []Attestor, opts ...AttestationContextOption) (*Attest
 }
 
 func (ctx *AttestationContext) RunAttestors() error {
-	preAttestors := []Attestor{}
-	materialAttestors := []Attestor{}
-	exeucteAttestors := []Attestor{}
-	productAttestors := []Attestor{}
-	postAttestors := []Attestor{}
-
+	attestors := make(map[RunType][]Attestor)
 	for _, attestor := range ctx.attestors {
-		switch attestor.RunType() {
-		case PreMaterialRunType:
-			preAttestors = append(preAttestors, attestor)
-
-		case MaterialRunType:
-			materialAttestors = append(materialAttestors, attestor)
-
-		case ExecuteRunType:
-			exeucteAttestors = append(exeucteAttestors, attestor)
-
-		case ProductRunType:
-			productAttestors = append(productAttestors, attestor)
-
-		case PostProductRunType:
-			postAttestors = append(postAttestors, attestor)
-
-		default:
+		if attestor.RunType() == "" {
 			return ErrInvalidOption{
-				Option: "attestor.RunType",
+				Option: "RunType",
 				Reason: fmt.Sprintf("unknown run type %v", attestor.RunType()),
 			}
 		}
 	}
 
-	for _, attestor := range preAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
-		}
-	}
-
-	for _, attestor := range materialAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
-		}
-	}
-
-	for _, attestor := range exeucteAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
-		}
-	}
-
-	for _, attestor := range productAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
-		}
-	}
-
-	for _, attestor := range postAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
+	for _, atts := range attestors {
+		for _, att := range atts {
+			log.Infof("Starting %v attestor...", att.Name())
+			if err := ctx.runAttestor(att); err != nil {
+				log.Errorf("Error running %v attestor: %w", att.Name(), err)
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


attestation/context.go
```diff
@@ -117,75 +117,35 @@ func NewContext(attestors []Attestor, opts ...AttestationContextOption) (*Attest
...
+		attestors[attestor.RunType()] = append(attestors[attestor.RunType()], attestor)
 	}
 
-	for _, attestor := range postAttestors {
-		if err := ctx.runAttestor(attestor); err != nil {
-			return err
+	for _, atts := range attestors {
+		for _, att := range atts {
+			log.Infof("Starting %v attestor...", att.Name())
+			if err := ctx.runAttestor(att); err != nil {
```
jkjell, line 135: It's not clear to me how this ensures attestors are run during their appropriate life cycle. 🤔

ChaosInTheCRD: hmmm I don't have an answer for you here. I can't recall seeing anything in the original logic that checked for this either, but i'll swing back and check this out.

mikhailswift: The way the original logic handled it was by first going through all the attestors and putting them into a slice specific to their lifecycle.

Then, we loop over each of these slices in order of the lifecycle stage and run each attestor in that stage's slice.

It's verbose, but simple and easy to grok at a glance :).

attestation/context.go
```diff
@@ -205,48 +165,34 @@ func (ctx *AttestationContext) runAttestor(attestor Attestor) error {
...
 	}
 
 	return nil
 }
 
 func (ctx *AttestationContext) CompletedAttestors() []CompletedAttestor {
-	attestors := make([]CompletedAttestor, len(ctx.completedAttestors))
-	copy(attestors, ctx.completedAttestors)
-	return attestors
+	return ctx.completedAttestors
```
mikhailswift, line 176: The thought process behind the `copy`s of this slice was to prevent someone from mucking about with the AttestationContext's internal state. For instance, by returning this my question is if we can now inadvertently add another Attestor to the backing array of `ctx.completedAttestors`

However, upon playing with this in the playground, it looks like these copys are entirely unnecessary and I was overthinking it  :) https://go.dev/play/p/iP1efJFDgzG

attestation/factory.go
```diff
@@ -41,6 +41,8 @@ type Subjecter interface {
 	Subjects() map[string]cryptoutil.DigestSet
 }
 
+// NOTE: not sure on the name of this interface, however I can't think of an alternative for now
```
mikhailswift, line 44: The weird world of Go interface naming https://go.dev/doc/effective_go#interface-names

run.go
```diff
@@ -85,6 +86,18 @@ func Run(stepName string, signer cryptoutil.Signer, opts ...RunOption) (RunResul
 		return result, fmt.Errorf("failed to run attestors: %w", err)
 	}
 
+	fail := false
+	for _, r := range runCtx.CompletedAttestors() {
```
mikhailswift, line 90: We can use `errors.Join` to wrap multiple errors here. Could look something like

```go
errors := make([]error, 0)
for _, r := range runCtx.CompletedAttestor() {
    if r.Error != nil {
        errors = append(errors, r.Error)
    }
}

if len(errors) > 0 {
    return result, errors.Join(errors.New("attestors failed with errors"), errors...)
}
```

attestation/context.go
```diff
@@ -205,48 +165,34 @@ func (ctx *AttestationContext) runAttestor(attestor Attestor) error {
...
 }
 
 func (ctx *AttestationContext) Materials() map[string]cryptoutil.DigestSet {
-	matCopy := make(map[string]cryptoutil.DigestSet)
-	for k, v := range ctx.materials {
-		matCopy[k] = v
-	}
-
-	return matCopy
+	return ctx.materials
```
mikhailswift, line 192: In contract to me overthinking it when it came to slices, it turns out I wasn't over thinking it when it came to maps. By returning the map, users can modify the unexported map within the struct by modifying the return value of this function:

https://go.dev/play/p/Ga5WdxpE9s4

In this case, we probably still want to copy the map before returning it.

attestation/context.go
mikhailswift: Ope, actually we should still copy these before returning. While appending doesn't affect the array in the struct, you can directly replace elements in the slice:

https://go.dev/play/p/ioXYWyituKW

attestation/context.go
ChaosInTheCRD: I'll have another look and see if we can marry the two in a way that's slimmer but still clear

mikhailswift (dismissed):
I like the improvement to the RunAttestors function in your latest changes :)

mikhailswift:
I'll take a peak at the linting job and see if I can sort anything out...

mikhailswift:
I cleared the golang-lint-ci cache for this job, and that cleared up the cache restoration error on this branch. It seemed to be an issue limited to this branch but if it crops up again we may need to take a deeper look at what's going on.

The lining is now erroring out because `errors.Join` doesn't exist in go 1.19, which our linting job is using currently due to that being what's set in our go.mod file. https://github.com/in-toto/go-witness/pull/146/commits/f976b77b3cb8d4dbad091db7775488a5adc91f91 updates this, you can cherry-pick it into this branch or re-create the commit by bumping the version and running a `go mod tidy`

ChaosInTheCRD:
I raised https://github.com/in-toto/go-witness/pull/145 to disable the cache and remove this issue going forward, but I am happy to leave that for now as you've sorted the problem. We can always make this change in the future if this reoccurs. Without sounding too ignorant do you mind me asking how you cleared the cach in the job @mikhailswift? 😆

ChaosInTheCRD:
On the subject of go version, this seems like a sensible time to bump https://github.com/in-toto/witness also. I think it would also be a good idea to figure out if there is any consideration we need to make in synchronizing the move to newer versions of go across repositories, and also some consideration on how long we wait before moving to new major versions. Just a thought.

mikhailswift:
> I raised #145 to disable the cache and remove this issue going forward, but I am happy to leave that for now as you've sorted the problem. We can always make this change in the future if this reoccurs. Without sounding too ignorant do you mind me asking how you cleared the cach in the job @mikhailswift? 😆

Yup! It's in the Actions section of the repo:

image

mikhailswift (approved):

PR #132, opened 2024-01-22
Moving the timestamper interfaces to the timestamp directory

ChaosInTheCRD:
This might not be an improvement we want to accept but it felt worthwhile moving the Timestamper interface(s) and the fake timestamper for testing to the `timestamp` package.

This is a bit of a "take or leave" PR, so happy for it to be closed if we don't feel it as necessary / optimal.

### DIFF
--- a/dsse/dsse_test.go
+++ b/dsse/dsse_test.go
@@ -16,19 +16,17 @@ package dsse
 
 import (
 	"bytes"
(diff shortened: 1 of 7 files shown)


mikhailswift (approved):
Yep, not sure why they ever lived in the dsse library 🤔

PR #139, opened 2024-01-29
refactor: move gitoid code to cyrptoutil, use digestvalue everywhere

mikhailswift:
When the functionality to calculate gitoids was added, there was a bit of tech debt incurred since they didn't implement hash.Hash. This remedies this with an admitedly hacky implementation of hash.Hash that wraps the gitoid code. This also standardizes our cryptoutil fucntions around the DigestValue struct that was added around this time to differentiate between gitoids and regular hash functions.

### DIFF
--- a/attestation/aws-iid/aws-iid.go
+++ b/attestation/aws-iid/aws-iid.go
@@ -80,7 +80,7 @@ func init() {
 
 type Attestor struct {
 	ec2metadata.EC2InstanceIdentityDocument
-	hashes    []crypto.Hash
+	hashes    []cryptoutil.DigestValue
 	session   session.Session
 	conf      *aws.Config
 	RawIID    string `json:"rawiid"`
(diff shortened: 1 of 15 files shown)


mikhailswift:
This is the first part of breaking up the VSA PR into a few smaller, easier to digest PRs.

ChaosInTheCRD (commented):
sorry for some reason it's not letting me comment inline - at [attestation/context.go:107](https://github.com/in-toto/go-witness/pull/139/files#diff-b7010efa1a357552dc2e67bdba448101161ca0f48500abf99b58e453ee8e96f3R107),  sorry for my ignorance, but why do we create 3 separate digest values?

ChaosInTheCRD:
This seems okay enough, tried my best to proof read and while I agree the solution is somewhat hacky but seems neater than how it was implemented before 👍

ChaosInTheCRD (approved):

ChaosInTheCRD:
I've approved, but won't merge as I have one small question. Don't think it's a blocker to merging though

mikhailswift:
> sorry for some reason it's not letting me comment inline - at [attestation/context.go:107](https://github.com/in-toto/go-witness/pull/139/files#diff-b7010efa1a357552dc2e67bdba448101161ca0f48500abf99b58e453ee8e96f3R107), sorry for my ignorance, but why do we create 3 separate digest values?

So these are the default hashes, and currently we were only using sha256 as our default. However, the material/product attestors were hard coded to also record sha256-gitoid and sha1-gitoid hashes, so they were added here to maintain consistent functionality

PR #144, opened 2024-01-29
Checking policy signature against cert constraints

ChaosInTheCRD:
This PR makes it possible to inspect the certificate attached to the signature (in the case that it was signed with x509 certificate) against user supplied cert constraints.

It should be noted that the constraint checks we perform is against *any* verifier that passes the initial policy verification logic. I am not sure whether this is an issue, but it is still something to be aware of. see [here](https://github.com/in-toto/go-witness/compare/main...ChaosInTheCRD:checking-policy-against-cert-constraints?expand=1#diff-32870c78c32d49e31ad4f22eaeac957ae6b988eea6fc937c9dba68035ac0d1b6R202-R212).

Another thing to note is that currently this logic has been broken out into a `verifyPolicySignature` function underneath the main `Verify` function. There may be a good place in the library to put this, but I couldn't think of one while working on it. Any suggestions on where we should put this would be most welcome.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -236,35 +236,13 @@ func (step Step) checkFunctionaries(verifiedStatements []source.VerifiedCollecti
 		}
 
 		for _, verifier := range verifiedStatement.Verifiers {
-			verifierID, err := verifier.KeyID()
-			if err != nil {
-				log.Debugf("(policy) skipping verifier: could not get key id: %w", err)
-				continue
-			}
-
 			for _, functionary := range step.Functionaries {
-				if functionary.PublicKeyID != "" && functionary.PublicKeyID == verifierID {
+				if err := functionary.Validate(verifier, trustBundles); err != nil {
+					log.Debugf("(policy) skipping verifier: %w", err)
+					continue
+				} else {
 					collections = append(collections, verifiedStatement)
-					break
 				}
-
-				x509Verifier, ok := verifier.(*cryptoutil.X509Verifier)
-				if !ok {
-					log.Debugf("(policy) skipping verifier: verifier with ID %v is not a public key verifier or a x509 verifier", verifierID)
-					continue
-				}
-
-				if len(functionary.CertConstraint.Roots) == 0 {
-					log.Debugf("(policy) skipping verifier: verifier with ID %v is an x509 verifier, but step %v does not have any truested roots", verifierID, step)
-					continue
-				}
-
-				if err := functionary.CertConstraint.Check(x509Verifier, trustBundles); err != nil {
-					log.Debugf("(policy) skipping verifier: verifier with ID %v doesn't meet certificate constraint: %w", verifierID, err)
-					continue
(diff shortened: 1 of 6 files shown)


mikhailswift (dismissed):
These changes look reasonable. I like the refactoring of breaking Functionary.Validate into it's own function.

policy/policy.go
```diff
@@ -236,35 +236,13 @@ func (step Step) checkFunctionaries(verifiedStatements []source.VerifiedCollecti
...
-				}
-
-				if len(functionary.CertConstraint.Roots) == 0 {
-					log.Debugf("(policy) skipping verifier: verifier with ID %v is an x509 verifier, but step %v does not have any truested roots", verifierID, step)
-					continue
-				}
-
-				if err := functionary.CertConstraint.Check(x509Verifier, trustBundles); err != nil {
-					log.Debugf("(policy) skipping verifier: verifier with ID %v doesn't meet certificate constraint: %w", verifierID, err)
+				if err := functionary.Validate(verifier, trustBundles); err != nil {
```
colek42, line 240: ⭐

policy/step.go
```diff
@@ -80,6 +81,32 @@ type RejectedCollection struct {
 	Reason     error
 }
 
+func (f Functionary) Validate(verifier cryptoutil.Verifier, trustBundles map[string]TrustBundle) error {
```
colek42, line 84: Agree with Mikhail, this is a great refactor

verify.go
```diff
@@ -85,6 +92,16 @@ func VerifyWithPolicyCAIntermediates(intermediates []*x509.Certificate) VerifyOp
 	}
 }
 
+func VerifyWithPolicyCertConstraints(commonName string, dnsNames []string, emails []string, organizations []string, uris []string) VerifyOption {
```
colek42, line 95: Do we want a test for this?

verify.go
```diff
@@ -154,3 +173,61 @@ func Verify(ctx context.Context, policyEnvelope dsse.Envelope, policyVerifiers [
 
 	return accepted, nil
 }
+
+func verifyPolicySignature(ctx context.Context, vo verifyOptions) error {
```
ChaosInTheCRD, line 177: I put this here, but it feels as though it is the wrong place for it.

Do you think this is an overthought or is there a good place to put this function? @mikhailswift @colek42.

verify.go
ChaosInTheCRD: yeah good idea

ChaosInTheCRD: Added tests 😄

ChaosInTheCRD:
@mikhailswift looks like we're getting the golangci-lint issue again 🤦

verify.go
jkjell: With the use of `verifyOptions`, this seems like a reasonable place.

jkjell (approved):
❤️

PR #149, opened 2024-02-05
Link & SLSA attestor

jkjell:
Add a new link attestor and the ability to return multiple signed attestation with `RunWithExports` function.


attestation/link/link.go
```diff
@@ -0,0 +1,154 @@
...
+	return RunType
+}
+
+func (l *Link) Export() bool {
+	return l.export
+}
+
+func (l *Link) Attest(ctx *attestation.AttestationContext) error {
+	l.PbLink.Name = "stepNameHere"
+	for _, attestor := range ctx.CompletedAttestors() {
```
ChaosInTheCRD, line 103: So this is less a comment specific to this code, but more something i'm wondering more generally:

> Should an attestor be able to inspect the metadata generated by other attestors? Even if the attestor is native to the tool (not a custom attestor like I have proposed elsehwere), is it _safe_ to attestors to basically interact with each others data?

ChaosInTheCRD, line 103: Am I right in saying that this *must* run last? I know we're running this as a "PostProduct", but could we run the risk of not catching other "PostProduct" attestors? Maybe this functionality would be better served as a flag that generates a link attestation rather than an attestor type?

jkjell: I was thinking that a new lifecycle type of "composite" that runs after all the other attestors. The composite attestor could be given a copy of the data instead of direct references. That wouldn't allow someone to "accidentally" modify the results of any other attestor.

jkjell: https://github.com/in-toto/go-witness/pull/149/#discussion_r1478435889

run.go
```diff
@@ -60,9 +61,24 @@ func RunWithTimestampers(ts ...timestamp.Timestamper) RunOption {
...
+// Deprecated: Use RunWithExports instead
 func Run(stepName string, signer cryptoutil.Signer, opts ...RunOption) (RunResult, error) {
+	results, err := run(stepName, signer, opts)
+	if len(results) > 1 {
+		return RunResult{}, errors.New("expected a single result, got multiple")
+	}
+	return results[0], err
+}
+
+func RunWithExports(stepName string, signer cryptoutil.Signer, opts ...RunOption) ([]RunResult, error) {
```
ChaosInTheCRD, line 77: Maybe I am not seeing the link attestor the right way, but to me turning `RunResult` into a slice of results is a bit confusing. It implies in my head that `Run` has been executed more than once. Instead I think I am right in saying that the link predicate is more of a way of packaging up references to the attestations? Given this, wouldn't it make more sense to return the link attestor as a field in the `RunResult` struct?

run.go
```diff
@@ -120,13 +152,13 @@ func validateRunOpts(ro runOptions) error {
 	return nil
 }
 
-func signCollection(collection attestation.Collection, opts ...dsse.SignOption) (dsse.Envelope, error) {
+func createAndSignEnvelope(collection interface{}, predType string, subjects map[string]cryptoutil.DigestSet, opts ...dsse.SignOption) (dsse.Envelope, error) {
```
ChaosInTheCRD, line 155: I think maybe keeping `signCollection` as is could be the right way here, and then maybe break out the contents to another shared function? 🤷 at the same time this might just be me getting the heebie jeebies by the fact that it still says `collection` for the input var. If this was change to something like `predicate`, maybe it would make more sense?

attestation/link/link.go
ChaosInTheCRD: Yeah I definitely think something that handled stuff **after** the attestors run would be the way to go here. It also keeps things decoupled as you say, to in the future completely lock attestors out from seeing each others metadata.

ChaosInTheCRD:
some opinionated thoughts from me, one little style thing that I think could be a tiny bit neater, but on the whole I ike the principle of the link connector and providing this way of exposing attestations. Code is also really neat and tidy 😄.

attestation/link/link.go
ChaosInTheCRD: Was this meant to another PR? this is just taking me to the top of this PR 😃 .

run.go
jkjell: I have no strong opinions about the naming and structure of the result. The important part is this is the boundary for signing envelopes. I'm actually not sure why RunResult has the `attestation.Collection` either. 🤔

attestation/link/link.go
colek42: We need to inspect to output from other attestors.  For example in the OCI attestor we query the product attestor for candidates.  It is the reason why we needed to create the lifecycle

colek42: would it make sense for the link attester to be post product?  I'm concerned about adding more steps to the lifecycle, and making the user define the order would not be ideal

jkjell: > would it make sense for the link attester to be post product
We actually need it to be the last post product. Example would be to add a subject for an OCI image. We have to run after the OCI post product attestor.

run.go
jkjell: Yeah, renaming to `predicate` makes sense.

attestation/link/link.go
ChaosInTheCRD: Seems it might be sensible to (at least for now) ensure that this attestor is handled last. I am also guessing that this sort of behaviour would need to apply to the SLSA attestor that sits in this PR. 

We should probably think more about the seemingly increasing number of attestors that will not gather information themselves but instead rely on evidence generated by others. Do we have an existing issue for this @jkjell @mikhailswift ?

ChaosInTheCRD: created one https://github.com/in-toto/witness/issues/422. Don't think we need to implement this in order to have this stuff merged though.

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (p *Provenance) Attest(ctx *attestation.AttestationContext) error {
+	builder := prov.Builder{}
+	metadata := prov.BuildMetadata{}
+	p.PbProvenance.BuildDefinition = &prov.BuildDefinition{}
+	p.PbProvenance.RunDetails = &prov.RunDetails{Builder: &builder, Metadata: &metadata}
+
+	p.PbProvenance.BuildDefinition.BuildType = "https://witness.dev/slsa-build@v0.1"
+	p.PbProvenance.RunDetails.Builder.Id = "https://witness.dev/witness-github-action@v0.1"
```
ChaosInTheCRD, line 112: @jkjell I think i'm right in saying that we need to add extra custom logic so that we can add this metadata depending on where / how witness was invocated? Maybe something that we might need to think about.

The way I see it, we should (as a first pass) look for attestors such as github, gitlab, google cloud build? If those attestors have run and are populated, we can use those to populate these fields.

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,199 @@
...
+	p.PbProvenance.RunDetails.Metadata.InvocationId = "gha-workflow-ref"
+
+	internalParamaters := make(map[string]interface{})
```
ChaosInTheCRD, line 115:
```suggestion
	internalParameters := make(map[string]interface{})
```

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+
+		case environment.Name:
+			envs := attestor.Attestor.(*environment.Attestor).Variables
+			pbEnvs := make(map[string]interface{}, len(envs))
+			for name, value := range envs {
+				pbEnvs[name] = value
+			}
+
+			internalParamaters["env"] = pbEnvs
```
ChaosInTheCRD, line 164:
```suggestion
			internalParameters["env"] = pbEnvs
```

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,199 @@
...
+
+		case product.ProductName:
+			p.products = attestor.Attestor.(*product.Attestor).Products()
+		}
+	}
+
+	var err error
+	p.PbProvenance.BuildDefinition.InternalParameters, err = structpb.NewStruct(internalParamaters)
```
ChaosInTheCRD, line 172:
```suggestion
	p.PbProvenance.BuildDefinition.InternalParameters, err = structpb.NewStruct(internalParameters)
```

ChaosInTheCRD: Could we put this in the environment attestor case statement?

run.go
ChaosInTheCRD: Yeah I was just looking at this. The `attestation.Collection` can be extracted from the `dsse.Envelope`, we could have a function attached to the RunResult type.

attestation/slsa/slsa.go
jkjell: There could be a few other internalParameters to add from the other attestors.

attestation/slsa/slsa.go
jkjell: Yep, GitHub and GitLab are the two we'll implement for sure. We'll have to come up with a reasonable name for a default or "undetected".

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,241 @@
...
+			if err != nil {
+				return err
+			}
+			// We have start and finish time at the collection level, how do we access it here?
+			p.PbProvenance.RunDetails.Metadata.StartedOn = timestamppb.New(time.Now())
+			p.PbProvenance.RunDetails.Metadata.FinishedOn = timestamppb.New(time.Now())
+
+		// Product Attestors
+		case product.ProductName:
+			maps.Copy(p.products, attestor.Attestor.(product.ProductAttestor).Products())
```
ChaosInTheCRD, line 205: hey @jkjell ! I pulled this down to see if I can make some QOL improvements, and I noticed that I was getting an `assignment to entry in nil map` error. Is this code running for you? I had to modify the code to initialise `p.products` in order to get it to work :)

ChaosInTheCRD: also would it make sense to leverage ctx.Products() here?

attestation/slsa/slsa.go
```diff
@@ -0,0 +1,241 @@
...
+
+		// Post Attestors
+		case oci.Name:
```
ChaosInTheCRD, line 208: - I see that we have the ProductAttestor interface https://github.com/in-toto/go-witness/blob/74a58da23ce6d08323d763807748d7584bad8ade/attestation/product/product.go#L50
- we cast OCI attestor to this interface if it is found:
```
		case oci.Name:
			maps.Copy(p.products, attestor.Attestor.(product.ProductAttestor).Products())
		}
```
- However, from what I can see the Products() function isn't implemented for this attestor :thinking_face:? https://github.com/in-toto/go-witness/blob/link-attestor/attestation/oci/oci.go

jkjell: Yeah, so this actually makes sense now that I think about it. The OCI attestor doesn't actually create any files but, it does create additional subjects for the OCI references it detects. I'll need to think through what subjects we actually want to add to the SLSA provenance attestation. 🤔

run.go
mikhailswift: I'm operating off of memory here -- but the `dsse.Envelope` stores the Collection as json bytes. Having the Collection as part of the RunResult was to allow users of the library to not have to re-parse the bytes to get the thing we just made for them.

mikhailswift:
Did a quick read through of the code and this all looks reasonable. Will pull and play with it tomorrow.

ChaosInTheCRD (approved):

ChaosInTheCRD:
provisionally approving, as long as we can resolve https://github.com/in-toto/go-witness/pull/149#discussion_r1552115108

ChaosInTheCRD:
Just noticed as well, we need to fix DCO before merge 😄

attestation/slsa/slsa.go
jkjell: So, for now, we're just adding all of the subjects. When we make future improvements to the OCI attestor, we should revisit this.

ChaosInTheCRD (approved):

PR #173, opened 2024-02-27
fix: reset verifier each iteration while loading pub keys from policy

mikhailswift:
Ran into an issue testing a policy with multiple public key functionaries. I would continually get ErrKeyIDMismatch when loading the second key. Turns out the verifier variable was being persisted on each iteration and failing the nil check to reload the new verifier. Resetting this within the for loop fixes it.

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -25,8 +25,10 @@ import (
 	_ "github.com/in-toto/go-witness/attestation/github"
 	_ "github.com/in-toto/go-witness/attestation/gitlab"
 	_ "github.com/in-toto/go-witness/attestation/jwt"
+	_ "github.com/in-toto/go-witness/attestation/material"
 	_ "github.com/in-toto/go-witness/attestation/maven"
 	_ "github.com/in-toto/go-witness/attestation/oci"
+	_ "github.com/in-toto/go-witness/attestation/product"
 	_ "github.com/in-toto/go-witness/attestation/sarif"
 
 	// signer providers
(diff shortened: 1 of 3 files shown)


mikhailswift:
making some adjustments to these tests real quick

mikhailswift:
okay this is good now

mikhailswift:
Also, I snuck in the addition of the material and product attestors to imports.go as part of this, since it came up earlier.

policy/policy.go
```diff
@@ -57,10 +57,10 @@ type PublicKey struct {
 // PublicKeyVerifiers returns verifiers for each of the policy's embedded public keys grouped by the key's ID
 func (p Policy) PublicKeyVerifiers() (map[string]cryptoutil.Verifier, error) {
 	verifiers := make(map[string]cryptoutil.Verifier)
-	var verifier cryptoutil.Verifier
 	var err error
 
 	for _, key := range p.PublicKeys {
+		var verifier cryptoutil.Verifier
```
kriscoleman, line 63: nice

colek42 (approved):
LGTM!  This must have been a fun one to track down.

mikhailswift:
yep. you know it's a sneaky bug when it results in a 1 line code change but 80 lines of tests.

jkjell (approved):

imports.go
```diff
@@ -25,8 +25,10 @@ import (
 	_ "github.com/in-toto/go-witness/attestation/github"
 	_ "github.com/in-toto/go-witness/attestation/gitlab"
 	_ "github.com/in-toto/go-witness/attestation/jwt"
+	_ "github.com/in-toto/go-witness/attestation/material"
```
ChaosInTheCRD, line 28: ❤️

policy/policy.go
ChaosInTheCRD: Apologies if this was my mistake 👀

ChaosInTheCRD (approved):

PR #197, opened 2024-03-28
JSON Schemas for attestors with generation scripts

ChaosInTheCRD:


jkjell (commented):
Looks good and very straight forward. 👍 

Two things this made me think about when reviewing:
- How are nested attestors represented (i.e. GitHub/GitLab using the JWT attestor)?
- For the (near-term) future: How do we handle schema versioning?

There's a few things that broke from the VSA stuff (and it'll break again after the Link/SLSA stuff too 😅).

schemagen/schema.go
```diff
@@ -0,0 +1,65 @@
...
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/in-toto/go-witness/attestation"
+	// this feels like an annoying way of importing them as it will need edited for each attestor added
```
jkjell, line 24: If you import `github.com/in-toto/go-witness/witness` will it pick up [imports.go](https://github.com/in-toto/go-witness/blob/main/imports.go)?

ChaosInTheCRD: yeah just done this all good 😄

ChaosInTheCRD:
>     * How are nested attestors represented (i.e. GitHub/GitLab using the JWT attestor)?

@jkjell good catch, it is represented like:

```
      "properties": {
        "jwt": {
          "$ref": "#/$defs/Attestor"
        },
```

so yeah, the nesting seems to be an issue. I'm not sure what the `tm` correct way of fixing this is (besides doing a version of what I did for the SLSA, Link and VSA attestors.


>     *For the (near-term) future: How do we handle schema versioning?

Ah, of course we handle versioning with predicate type, but this isn't included in the JSON schema...

schemagen/material.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$ref": "#/$defs/Attestor",
+  "$defs": {
+    "Attestor": {
+      "properties": {},
```
jkjell, line 6: No properties for material attestor?

schemagen/product.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$ref": "#/$defs/Attestor",
+  "$defs": {
+    "Attestor": {
+      "properties": {},
```
jkjell, line 6: No properties for the product attestor?

jkjell (commented):
Everything looks good other than the material and product attestor generation. If those are resolved we'll be good to go. ✅

schemagen/material.json
ChaosInTheCRD: 🤦

jkjell:
> should I try and find a solution or do you think it's fine as is for now?

It's good for now. This is WAY better than what we have now... i.e. the source code. 😂

jkjell (approved):

colek42:
This is a great step forward.  Aligning with upstream using protos would be nice, but a lot more work

PR #209, opened 2024-04-11
feat: Add field `name` in the Policy

kairoaraujo:
Includes the field `name` for the Policy as an optional. The `name` helps users define and identify a policy with a unique identifier, allowing even versioning.

Motivation:
It can be difficult to manage and identify multiple policies by a file name or its content, depending on the storage usage.
In automation, this field can also be helpful.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -1,4 +1,4 @@
-// Copyright 2021 The Witness Contributors
+// Copyright 2021-2024 The Witness Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


policy/policy.go
```diff
@@ -36,6 +36,7 @@ const PolicyPredicate = "https://witness.testifysec.com/policy/v0.1"
 // +kubebuilder:object:generate=true
 type Policy struct {
 	Expires              metav1.Time          `json:"expires"`
+	Name                 string               `json:"name,omitempty"`
```
kairoaraujo, line 39: Should it have some constraints?
i.e. `([A-Za-z0-9\-\_]+)`

adityasaky:
I'm a bit nervous about this. On the one hand, it can be useful to ensure a supply chain owner _signs_ the right policy and the orchestration actually provides the right policy for verification. But at the same time, I'm worried about a policy self identifying itself (where only the name + supporting data in the policy is used during verification) without appropriately bootstrapping the policy metadata's trust. If we can't guarantee a policy is bootstrapped using ITE-2/3 semantics, we should maybe tread carefully?

jkjell:
I had a chat with @kairoaraujo about this today. I think we'll store the policy name in the Archivista metadata instead of within the policy itself.

kairoaraujo:
I misunderstood it. We don't require it in the Policy metadata.

PR #223, opened 2024-04-29
Adding functionality for dirhash in library

matglas:
Related to: https://github.com/in-toto/witness/pull/436

This a different approach to #65. It captures a dirhash for material and product if the dir is mentioned as an argument.

```
witness run --dirhash-glob node_modules/* --step build -- npm clean-install
```

The argument can be repeated multiple time and globs with `**` are not allowed.

The output will generate a dirhash based on the algorithm use by golang in https://pkg.go.dev/golang.org/x/mod/sumdb/dirhash and is the same as described in the in-toto digest spec ([here](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#dirhash))

The subject for a dirhash will be `https://witness.dev/attestations/product/v0.1/dir:node_modules/example/`.

The policy works correctly and also `artifactsFrom` are correctly checked.

## Todo

- [x] allow verify to check not only file subject but also dir subject
- [x] decide on the copying of the glob library into the go-witness library.
- [x] discuss what additional docs on the library would be benificial

### DIFF
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gobwas/glob"
 	"github.com/in-toto/go-witness/cryptoutil"
 	"github.com/in-toto/go-witness/log"
 )
@@ -83,6 +84,20 @@ func WithWorkingDir(workingDir string) AttestationContextOption {
 	}
 }
 
+func WithDirHashGlob(dirHashGlob []string) AttestationContextOption {
+	return func(ctx *AttestationContext) {
+		if len(dirHashGlob) > 0 {
+			ctx.dirHashGlob = dirHashGlob
+
+			ctx.dirHashGlobCompiled = make([]glob.Glob, len(ctx.dirHashGlob))
+			for i, dirHashGlobItem := range dirHashGlob {
+				dirHashGlobItemCompiled, _ := glob.Compile(dirHashGlobItem)
+				ctx.dirHashGlobCompiled[i] = dirHashGlobItemCompiled
+			}
+		}
+	}
+}
+
 type CompletedAttestor struct {
 	Attestor  Attestor
 	StartTime time.Time
@@ -91,15 +106,17 @@ type CompletedAttestor struct {
 }
 
 type AttestationContext struct {
-	ctx                context.Context
-	attestors          []Attestor
-	workingDir         string
-	hashes             []cryptoutil.DigestValue
-	completedAttestors []CompletedAttestor
-	products           map[string]Product
-	materials          map[string]cryptoutil.DigestSet
-	stepName           string
-	mutex              sync.RWMutex
+	ctx                 context.Context
+	attestors           []Attestor
+	workingDir          string
+	dirHashGlob         []string
+	dirHashGlobCompiled []glob.Glob
+	hashes              []cryptoutil.DigestValue
+	completedAttestors  []CompletedAttestor
+	products            map[string]Product
+	materials           map[string]cryptoutil.DigestSet
+	stepName            string
+	mutex               sync.RWMutex
 }
 
 type Product struct {
@@ -206,6 +223,10 @@ func (ctx *AttestationContext) runAttestor(attestor Attestor) {
 	}
 
 	log.Infof("Finished %v attestor... (%vs)", attestor.Name(), time.Since(startTime).Seconds())
+}
+
+func (ctx *AttestationContext) DirHashGlob() []glob.Glob {
+	return ctx.dirHashGlobCompiled
 }
 
 func (ctx *AttestationContext) CompletedAttestors() []CompletedAttestor {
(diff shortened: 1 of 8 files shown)


attestation/context.go
```diff
@@ -77,6 +78,20 @@ func WithWorkingDir(workingDir string) AttestationContextOption {
...
+			ctx.dirHashGlob = dirHashGlob
+
+			ctx.dirHashGlobCompiled = make([]glob.Glob, len(ctx.dirHashGlob))
+			for i, dirHashGlobItem := range dirHashGlob {
+				dirHashGlobItemCompiled, _ := glob.Compile(dirHashGlobItem)
+				ctx.dirHashGlobCompiled[i] = dirHashGlobItemCompiled
+			}
+		}
+	}
+}
```
matglas, line 93: This is to add a the dirhash to the global context so multiple attestations can use it. It would have been easier to do input validation here but design wise that was not possible.

I am doing the input validation now on the witness run.go side. But I did not create a PR for it yet.

attestation/file/file.go
```diff
@@ -19,29 +19,48 @@ import (
...
+
+			if dirHashMatch {
+				dir, _ := cryptoutil.CalculateDigestSetFromDir(path, hashes)
+
+				if err != nil {
+					return err
+				}
+
+				artifacts[relPath + string(os.PathSeparator)] = dir
+				return filepath.SkipDir
```
matglas, line 58: I'm skipping any further processing of files inside this dir.

attestation/product/product.go
```diff
@@ -107,11 +107,23 @@ func fromDigestMap(digestMap map[string]cryptoutil.DigestSet) map[string]attesta
...
+			}
+
+			// IsDir is short for fileInfo.Mode().IsDir()
+			if fileInfo.IsDir() {
+				mimeType = "text/directory"
+			} else {
+				mimeType, err = getFileContentType(f)
+				if err != nil {
+					mimeType = "unknown"
+				}
```
matglas, line 126: We'll check here if its a directory this allows later processing too into the subject output. And its possible to understand the digest. The mime-type is a known mime-type but not actually true in the sense that we do not have mime-type data here about the folder. But it works quite practically.

attestation/product/product.go
```diff
@@ -202,7 +214,11 @@ func (a *Attestor) Subjects() map[string]cryptoutil.DigestSet {
 			continue
 		}
 
-		subjects[fmt.Sprintf("file:%v", productName)] = product.Digest
+		subjectType := "file"
+		if product.MimeType == "text/directory" {
+			subjectType = "dir"
+		}
+		subjects[fmt.Sprintf("%v:%v", subjectType, productName)] = product.Digest
```
matglas, line 221: Here we are using the mime-type we have set to be able to give our subject a different prefix.

cryptoutil/dirhash.go
```diff
@@ -0,0 +1,47 @@
...
+//
+//	sha256sum $(find . -type f | sort) | sha256sum
+//
+// More precisely, the hashed summary contains a single line for each file in the list,
+// ordered by sort.Strings applied to the file names, where each line consists of
+// the hexadecimal SHA-256 hash of the file content,
+// two spaces (U+0020), the file name, and a newline (U+000A).
+//
+// File names with newlines (U+000A) are disallowed.
+func DirhHashSha256(files []string, open func(string) (io.ReadCloser, error)) (string, error) {
```
matglas, line 26: This is a copy of the Hash1 function with the twist that is does not return a `h1:` string but the original sha256 hash.

ChaosInTheCRD:
Hey! thanks for submitting this 😄. It looks like this is the sort of functionality we were looking for after the discussion on the community call, so thanks a lot for following up with it.

After looking through things, it seems as though everything looks good, although the use of "github.com/gobwas/glob" worries me a tad since it's a package that belongs to an inactive user and hasn't been committed to for 6 years.

It's probably a tad ironic since at this time we don't have any strong policy over the kind of packages we are and are not happy to pull in. But maybe in this case, a bit of copying might be better than depending? 🤔 Just a thought.

I would add also that we need to get tests written before this is merged, and docs updated in github.com/in-toto/witness before https://github.com/in-toto/witness/pull/436 is merged.

matglas:
Thanks for the input. What is ok OSS wise in copying code?

adityasaky:
I haven't had a chance to look at this PR, but I wanted to check if it's consistent with the implementation in in-toto/in-toto. Is that the case? Thanks!

matglas:
@ChaosInTheCRD I updated and fixed the tests to do some dirhash testing.

matglas:
Cherry-picked all commits onto latest main to fix DCO issues.

matglas:
@jkjell all should be good to merge now. There are not breaking changes to witness. The usage will be made available in the other repo. But even with the changes in it should not break any other behavior.

jkjell (approved):
Great work @matglas! 🎉

matglas:
Great! Thank you!

PR #228, opened 2024-04-30
Parallel attestors per type

matglas:
Allowing per attestor type to run all the attestors in parallel to speed up the process.

- [x] Check for need of mutex in different attestors when using data from context.

### DIFF
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/in-toto/go-witness/cryptoutil"
@@ -98,6 +99,7 @@ type AttestationContext struct {
 	products           map[string]Product
 	materials          map[string]cryptoutil.DigestSet
 	stepName           string
+	mutex              sync.RWMutex
 }
 
 type Product struct {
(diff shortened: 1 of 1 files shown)


ChaosInTheCRD:
I seem to recall that there was a reason why we opted to not run attestors in parallel 🤔 @mikhailswift @jkjell any ideas?

attestation/context.go
```diff
@@ -138,16 +139,27 @@ func (ctx *AttestationContext) RunAttestors() error {
 	order := runTypeOrder()
 	for _, k := range order {
 		log.Debugf("Starting %s attestors...", k.String())
```
matglas, line 141: Thinking of adjusting the log line here to make it clear that this is the start of a phase. And maybe add a Name method to OrderType so we can give a better visual feedback.

attestation/context.go
```diff
@@ -171,6 +183,8 @@ func (ctx *AttestationContext) runAttestor(attestor Attestor) {
 	if producer, ok := attestor.(Producer); ok {
 		ctx.addProducts(producer)
 	}
+
+	log.Debugf("Finished %v attestor...", attestor.Name())
```
matglas, line 187: Thinking of making this a `Infof` instead of `Debugf` so you get more understandable output.

jkjell (approved):
Given the scope of the changes, I'll wait for a second approval from @ChaosInTheCRD or @mikhailswift before merging.

matglas:
Nice thanks!

PR #230, opened 2024-05-01
Proposed changes for main verification attestor PR

ChaosInTheCRD:
These are the implementations of the proposals made in the verification attestor PR, mainly to tidy things up a bit and ensure a good experience in https://github.com/in-toto/witness.

@mikhailswift feel free to accept / reject these proposals as you see fit 😄.


attestation/context.go
```diff
@@ -33,6 +33,7 @@ const (
 	ExecuteRunType     RunType = "execute"
 	ProductRunType     RunType = "product"
 	PostProductRunType RunType = "postproduct"
+	VerifyRunType      RunType = "verify"
```
ChaosInTheCRD, line 36: I've changed `policyverify` to to have a `VerifyRunType`. Currently, you can actually perform `run -a=policyverify`, which of course fails as it isn't correctly scaffolded in the CLI. By adding this run type there is logic (here) that lets us ensure that it isn't used in tandem with other types.

attestation/policyverify/policyverify.go
ChaosInTheCRD, line 1: Not sure why this is showing as a completely new file?

jkjell (approved):
We can make further changes in the branch off `in-toto/go-witness` as needed.

PR #257, opened 2024-05-23
Step analyze fix

jkjell:
If any evidence exists that causes a step evaluation to fail, the entire policy will fail, even if there's passing evidence at another search depth. Also, results from the verifyArtifacts function are carrying forward passed results, even if the artifact comparison doesn't match.


kriscoleman (approved):
👍 good catch

ChaosInTheCRD (approved):

ChaosInTheCRD:
Sorry for leaving this for so long

mikhailswift (approved):

PR #260, opened 2024-05-30
feat: add ability to sign with vault transit secret engine

mikhailswift:


colek42:
My recommendation:

1:  Save the full signature exactly as Vault returns it: vault:v1:BASE64ENCODEDSIGNATURE.
2:  When verifying locally (with the local public key), strip off vault:v1:.
3:  When verifying remotely (with the Transit engine), pass the signature through unchanged, which includes vault:v1:.

mikhailswift:
No

rabajaj0509:
@mikhailswift, merging this would be very helpful, as I have a use case for this PR :)

mikhailswift:
This is ready for a review. It's been used in practice for a while and works well.

colek42 (dismissed):
I'd like to see a test standing up open source vault (or a mock).  We need test for both online and offline verification.

colek42:
> > I'd like to see a test standing up open source vault (or a mock). We need test for both online and offline verification.
> 
> Tests are being worked on. Going to mock it, don't think standing up an instance of vault is worth the complexity.

FWIW, I was able to test this today using the open source Vault server.  Online verification worked great, but had some issues with offline.  However I think offline verification with transit KMS is something that is unsolved across the entire code base.

colek42:
> > > > I'd like to see a test standing up open source vault (or a mock). We need test for both online and offline verification.
> > > 
> > > 
> > > Tests are being worked on. Going to mock it, don't think standing up an instance of vault is worth the complexity.
> > 
> > 
> > FWIW, I was able to test this today using the open source Vault server. Online verification worked great, but had some issues with offline. However I think offline verification with transit KMS is something that is unsolved across the entire code base.
> 
> Honestly rethinking my opinion that mocking it is the way to go. Something like testcontainers and spinning up a vault server may honestly be easier and a more meaningful test. I just don't like having the basic `go test ./...` require external dependencies or to be slow to run. Could use tags to separate integration tests that depend on external services with something like
> 
> ```go
> // +build integration
> ```
> 
> and then run enable them with `go test -tags=integration`
> 
> Agree that offline verification is something that needs some attention across all KMS signers. I think there's some half baked functionality for AWS to support offline verification. It's probably worth giving that some attention across all the signers.

This is what we do internally.  Vault is pretty lightweight.

PR #291, opened 2024-06-24
Define more than one `Type` per Attestor

ChaosInTheCRD:

This PR introduces the ability to define more than one Predicate Type through the `Type` method in the `Attestor` interface. The first string in the `TypeSet` is treated as the "primary" type (e.g., `https://witness.dev/attestations/sbom/v0.1`) and can be returned with the `First()` function.

When initializing the `attestationsByType` map, each type in the `TypeSet` for a given attestor is added as a key with the attestor as its value.


ChaosInTheCRD:
Closing this as it is a duplicate of work performed by @jkjell

PR #292, opened 2024-06-24
fix: Passing kms provider options down to initialisation of functionaries

ChaosInTheCRD:

this closes https://github.com/in-toto/witness/issues/427

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -43,4 +43,4 @@ jobs:
         with:
           version: latest
           args: --timeout=3m
-          skip-pkg-cache: true
+          skip-cache: true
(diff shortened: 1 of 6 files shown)


.github/workflows/golangci-lint.yml
ChaosInTheCRD, line 1: Changed this field as I was getting:

```
Warning: Unexpected input(s) 'skip-pkg-cache', valid inputs are ['version', 'install-mode', 'working-directory', 'github-token', 'only-new-issues', 'skip-cache', 'skip-save-cache', 'problem-matchers', 'args', 'cache-invalidation-interval']
```

I think this was the correct modification? @jkjell

jkjell (approved):

PR #306, opened 2024-07-15
better concurrency support on Linux

joshdabosh:
- When ptracing programs, restrict the product attestor to only attest for files recorded as "opened" by the commandrun attestor. Then, files created by a separate process running concurrently will not be mistakenly included in the attestation products (this happens currently).
- Keep track of newly created files when tracing with ptrace. Previously, they would not be added to the `OpenedFiles` list because the file would not exist when attempting to calculate its digest.

## Acceptance Criteria Met

- [x] Docs changes if needed

### DIFF
--- a/attestation/commandrun/commandrun.go
+++ b/attestation/commandrun/commandrun.go
@@ -70,7 +70,7 @@ func WithMaterials(materials map[string]cryptoutil.DigestSet) Option {
 
 func WithTracing(enabled bool) Option {
 	return func(cr *CommandRun) {
-		cr.enableTracing = enabled
+		cr.EnableTracing = enabled
 	}
 }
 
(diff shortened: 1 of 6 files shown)


mikhailswift:
Hi @joshdabosh, thanks for this contribution! I'll start testing this and give you a review shortly.

attestation/commandrun/commandrun.go
```diff
@@ -160,6 +160,10 @@ func (rc *CommandRun) RunType() attestation.RunType {
 	return RunType
 }
 
+func (rc *CommandRun) EnableTracing() bool {
```
mikhailswift, line 163: This may be nitpicky, but could we rename this `TracingEnabled` instead? `EnableTracing` reads to me like "take the action to enable tracing" versus "ask if tracing is enabled".

mikhailswift:
Just the one minor nit, otherwise I think this is good to go!

attestation/commandrun/commandrun.go
joshdabosh: done!

mikhailswift (approved):

Issue #275, opened 2024-06-12
[Feat]: Extensibility for the environment attestor


joshdabosh:
**Describe the solution you'd like:**
A way to add environment variables to the default denylist, or a way to optionally only attest a specified list of environment variables.

**User value:**

Some environments may have sensitive information stored in environment variables which are not covered in the denylist now. Using the environment attestor will store them in recoverable plaintext in generated attestations.

**Expected behavior:**

Environment variables can be selectively denied or allowed to be attested.

**Proposed solution:**

Add a flag to decide between allowlist / denylist, and another flag to append to either list for later filtering.

**Anything else you would like to add:**

**Testing changes required:**

Add tests to make sure environment variables are correctly filtered.

**Documentation changes required:**

Add flags to the CLI reference and modify the Environment attestor page.

matglas:
This would be very valuable to have! A list that you add would be helpful. Also a more generic approach of excluding all `*_TOKEN` could be helpful too.

@jkjell maybe this could have a feature label.

jkjell:
One thing for future conversation about this is what to do with environmental variables that modify the behavior of a process. If someone decides to compile a program with `ENABLE_INSECURE_BEHAVIOR=1` but, has a `DENY *` config for the environment attestor, a policy could still need to be able to verify the presence or absence of that flag. This is a long way of saying we'll probably need to include the config options of the attestor in the attestation itself.

matglas:
> we'll probably need to include the config options of the attestor in the attestation itself.

Agree. I think this is something that all attestors should have.

matglas:
Would a PR that protects vars by hidding the value be a good first step?

matglas:
@joshdabosh please see the PR and the conversation that is happening there. Curious to hear your thoughts on proposed enhancements there.

matglas:
@joshdabosh the new functionality is merged and released in v0.8.0. Please give it a try.

joshdabosh:
Hey - looks good to me, thanks!

PR #355, opened 2024-09-20
feat: New environment variable obfuscation functionality

matglas:

Implementing obfuscation of environment variables. Capturing secret values like tokens and api keys is a security risk and attestation should not hold that information. Although it is important to know which values are used to adjust the behavior for scripts.

Fixes:
- https://github.com/in-toto/go-witness/issues/275

## Acceptance Criteria Met

- [x] Docs changes if needed

The default list can be extended with more variable that should be hidden. This can be extended in the cli with flags that add more items to the list. I think that the default obfuscation list does not need to be 'overridden' but only needs extension.

It also obfuscates based on glob. So if an key would be added like `REASON_*` it will pick up all `REASON_TEST` etc. variables. There are a few glob keys that are added by default.

```
*_TOKEN
SECRET_*
*_API_KEY
*_PASSWORD
*_JWT
```

**Final solution**

New flags:
- `--env-filter-sensitive-vars`
- `--env-disable-default-sensitive-vars`
- `--env-add-sensitive-key` can be repeated multiple times.
- `--env-exclude-sensitive-key` can be repeated multiple times.

Blocklist is now filter list. It describes better what it does.
Obfuscation is the default behavior.
Variables can be obfuscated or filtered based on glob patterns e.g. `*SECRET`

There is a default list of globs which is quite restrictive by default allow users to explicitly use `exclude-sensitive-key` to exclude it from filtering or obfuscation.

```
*TOKEN*
*SECRET*
*API_KEY*
*PASSWORD*
*JWT*
```


matglas:
@jkjell would you be able to do a review?

attestation/environment/obfuscate.go
```diff
@@ -0,0 +1,66 @@
...
+
+// sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
+func DefaultObfuscateList() map[string]struct{} {
+	return map[string]struct{}{
+		"*_TOKEN":    {},
+		"SECRET_*":   {},
+		"*_API_KEY":  {},
+		"*_PASSWORD": {},
+		"*_JWT":      {},
+	}
```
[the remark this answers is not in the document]
matglas: Context: Currently the block list completely removes those variables. The obfuscate will keep them but hide the content. 

I am not sure if combining those lists is a good option in that case.

About the wildcard. I can understand. I do think that this obfuscate defaults is just a 'primary' list. And users might need to extend it for their own cases. Actually many of the items in the block list 'could' be removed and be covered by the obfuscate.

As I am thinking about this now it might still be very valuable to extend the blocklist. Because you do not want to expose to consumers where you are using a specific token. 🤔 The obfuscate is basically a stop gap solution. So maybe you are right and we should obfuscate any `*<WORD>*` like `*TOKEN*`.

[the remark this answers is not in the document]
kairoaraujo: > I am not sure if combining those lists is a good option in that case.

I don't understand why the source list of blocking and obfuscating should be different, as both just have different behaviors when removing or obfuscating the data.

In other words. As a user, I expect them to differ on the "remove"/"obfuscate" my environment variables.
As a user, I shouldn't need to understand that actions uses different criteria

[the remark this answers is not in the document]
matglas: True. Obfuscate now is complementary to the blocklist without changing the current behavior. But in the end it would be better to use a flag that says e.g. "block and obfuscate by default".

matglas:
With this latest change I go from default blocking behavior to default obfuscate behavior. And the default list
is the same for blocking and obfuscating. Also the blocking will also use glob patterns now.

Introducing several arguments:

`--attestor-environment-block-sensitive-vars`  
Switch from obfuscate to blocking variables which removes them from the output completely.

`--attestor-environment-disable-default-sensitive-vars`  
Disable the default list of sensitive vars and only use the items mentioned by --attestor-environment-sensitive-key.

`--attestor-environment-sensitive-key` can be repeated  
Add keys to the list of sensitive environment keys.

I believe that as we are still in the v0.x.y versions it should not be a problem to change the default
behavior from blocking to obfuscating. The variables will still not be exposed and switching to blocking 
is still possible.

@mikhailswift I need a bit of help on the `--attestor-environment-sensitive-key` and it looks like you worked on the `registry.StringSliceConfigOption`. But its not capturing the values correctly somehow. Are you able to help me a bit here?

matglas:
Forget the part about the `registry.StringSliceConfigOption`. I had a misconfiguration in my debug setup. But I did found another missing part after that 😄 . The function works now properly. It only needs more work on the tests.

matglas:
My last changes fixed the tests. I also renamed block to filter. Which seems to me that it is a much more logical.

attestation/environment/sensitive_env_vars.go
```diff
@@ -14,9 +14,20 @@
...
+// Eplicit list sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
+func DefaultSensitiveEnvList() map[string]struct{} {
 	return map[string]struct{}{
+
+		// Glob pattern list
+		"*_TOKEN":    {},
+		"SECRET_*":   {},
+		"*_API_KEY":  {},
+		"*_PASSWORD": {},
+		"*_JWT":      {},
```
kairoaraujo, line 28: We should make it less opinionated on how the env var should start or finish.
If you see, for secret, we are ok with having `SECRET_*`, but for token, we expect a prefix of `*_TOKEN`.

With the (amazing) last changes you added (exclude sensitive key), I think we could make it more strict

 
```suggestion
		"*TOKEN*":    {},
		"*SECRET*":   {},
		"*API_KEY*":  {},
		"*PASSWORD*": {},
		"*JWT*":      {},
```

matglas: Sounds good! Thanks for that addition.

kairoaraujo (dismissed):
LGTM.
Waiting for others feedback

matglas:
@ChaosInTheCRD and I had a conversation on Slack about the implementation because there is a 'hidden' use of Environment inside the commandrun attestor. It used to use the `FilterEnvironmentArray` but the behavior is completely changing because of the new flags. Because of this commandrun can not use it correctly anymore.

We discussed that is might be best to introduce the current new flags not on the environment attestor level, but move them to the global flags. This way the flags can be used both in Environment and Commandrun Attestor.

Also the filtering and obfuscation methods need to made generic for. So not specifically as part of the attestors.

@kairoaraujo @ChaosInTheCRD if we agree on this I'll implement the change that way. A 👍 is enough.

axi92:
Nice to see that this gets implemented! :+1: 
I asked about that in a discussion https://github.com/in-toto/witness/discussions/510
And I love to see the support for bamboo from Atlassian too.
Since the original list was from a separate repo I posted an issue there https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/issues/12. Maybe it could be implemented here as well?

axi92:
Yes, obfuscating all bamboo variables would not make sense. With the list https://github.com/matglas/go-witness/blob/ecacd4ec0ba7a4a12664d3e5d75eedbd433bb9f2/attestation/environment/sensitive_env_vars.go#L24-L28 of defaults are those case sensitive?

Secret and password are already covered right now.
But `sshKey` and `passphrase` is missing, if an env variable contains one of those two, they are obfuscated per default on bamboo.

attestation/environment/environment.go
```diff
@@ -47,9 +48,65 @@ type EnvironmentAttestor interface {
 }
 
 func init() {
-	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor {
-		return New()
-	})
+	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor { return New() })
+
+	// registry.BoolConfigOption(
```
ChaosInTheCRD, line 53: sorry if I'm missing some context, is there a reason why this is commented out at the moment @matglas ?

attestation/commandrun/tracing_linux.go
```diff
@@ -53,11 +53,11 @@ func enableTracing(c *exec.Cmd) {
 
 func (r *CommandRun) trace(c *exec.Cmd, actx *attestation.AttestationContext) ([]ProcessInfo, error) {
 	pctx := &ptraceContext{
-		parentPid:            c.Process.Pid,
-		mainProgram:          c.Path,
-		processes:            make(map[int]*ProcessInfo),
-		hash:                 actx.Hashes(),
-		environmentBlockList: r.environmentBlockList,
```
ChaosInTheCRD, line 60: Just as a sanity check, we're ensuring backwards compatibility here right? Again, been out of the loop on this last week or two but just asking the questions to ensure that this was all considered 😄

environment/sensitive_env_vars.go
```diff
@@ -14,9 +14,22 @@
 
 package environment
 
-// sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
-func DefaultBlockList() map[string]struct{} {
+// DefaultSensitiveEnvList return a list of known sensitive environment keys.
+//
+// Eplicit list sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
```
ChaosInTheCRD, line 19:
```suggestion
// Explicit list sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
```
```suggestion
// Explicit list sourced from https://github.com/Puliczek/awesome-list-of-secrets-in-environment-variables/blob/main/raw_list.txt
```

attestation/environment/environment.go
matglas: Its a WIP. This has to move to cli part to make it work again with flags.

attestation/commandrun/tracing_linux.go
matglas: Yes backward compatibility on the tracing. But with the side note that we now do not filter anymore but use configuration of the `EnvironmentCapturer`

attestation/environment/environment.go
matglas: This is fixed now in PR https://github.com/in-toto/witness/pull/522

attestation/commandrun/tracing_linux.go
matglas: If you don't see a problem with that we can resolve this conversation.

matglas:
@jkjell do you think we could include this in the next release too?

colek42 (approved):

matglas:
Thank you for merging these changes. How do we proceed to add this PR that adds the CLI part too?

See: https://github.com/in-toto/witness/pull/522

Issue #358, opened 2024-09-24
[Feat]: Speed up the git status part of the git attestor


matglas:
**Describe the solution you'd like:**

The current part that is capturing the git status in the git attestor is slow especially when using a large repository. Go-git does not do threading and capturing the status takes a long time because of that.

**User value:**

By improving the speed users with larger repositories will see a significant speed up in the execution time of witness. This plays a part in adoption. As a DevOps engineer I want to keep my pipelines fast but I do want to capture attestations with Witness.

**Expected behavior:**

Status should be captured as fast as doing a normal `git status` on a repo.

**Proposed solution:**

Replace go-git or use git cli as a hybrid solution to capture the status of the repo in the attestation.

**Testing changes required:**

TBD 

**Documentation changes required:**

TBD

matglas:
I think this might be an enhancement. But we can't set labels and this template worked the best 😄 
@jkjell can you adjust this?

PR #359, opened 2024-09-24
chore: Improve the git status speed.

matglas:

By using the native git client its possible to improve the speed significantly.
The default approach is using go-git to no rely on a git client that is installed. But if git is found it will use the git client instead which is much faster compared to the go-git library.

Fixes https://github.com/in-toto/go-witness/issues/358

## Acceptance Criteria Met

- [x] Docs changes if needed

I have large mono repo. For posterity here are the results of a simple run that does not change anything and then with the changes. Basically: `witness run --step build -- echo "hello"`

Repository stats of the working tree.
```shell
$ git count-objects -vH

count: 219
size: 1.13 MiB
in-pack: 2468342
packs: 29
size-pack: 1.36 GiB
prune-packable: 25
garbage: 0
size-garbage: 0 bytes
```

v0.6.0: 82.82s user 96.35s system 63% cpu 4:40.58 total
change: 55.27s user 66.71s system 66% cpu 3:02.30 total

### DIFF
--- a/attestation/git/git_bin.go
+++ b/attestation/git/git_bin.go
@@ -15,10 +15,13 @@
 package git
 
 import (
+	"fmt"
 	"os/exec"
 	"strings"
 
 	"github.com/go-git/go-git/v5"
+	"github.com/in-toto/go-witness/attestation"
+	"github.com/in-toto/go-witness/cryptoutil"
 )
 
 // GitExists checks if the git binary is available.
@@ -33,6 +36,40 @@ func GitExists() bool {
 	}
 }
 
+// GitGetBinPath retrieves the path to the git binary that is used by the attestor.
+func GitGetBinPath() (string, error) {
+	path, err := exec.LookPath("git")
+	if err != nil {
+		return "", err
+	} else {
+		return path, nil
+	}
+}
+
+// GitGetBinHash retrieves a sha256 hash of the git binary that is located on the system.
+// The path is determined based on exec.LookPath().
+func GitGetBinHash(ctx *attestation.AttestationContext) (string, error) {
+	path, err := exec.LookPath("git")
+	if err != nil {
+		return "", err
+	}
+
+	gitBinDigest, err := cryptoutil.CalculateDigestSetFromFile(path, ctx.Hashes())
+	fmt.Printf("%s", gitBinDigest)
+	if err != nil {
+		return "", err
+	}
+
+	digestMap, err := gitBinDigest.ToNameMap()
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("sha256:%s", digestMap["sha256"]), nil
+}
+
+// GitGetStatus retrieves the status of staging and worktree
+// from the git status --porcelain output
 func GitGetStatus(workDir string) (map[string]Status, error) {
 
 	// Execute the git status --porcelain command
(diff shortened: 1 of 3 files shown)


ChaosInTheCRD:
Hey @matglas!

This all looks good to me from a first look. Of course the question of switching from `go-git` to `git` is a question in of itself, one that I am sure that others will have opinions on.

What I will say is that I have seen the use of the `git` CLI inside go-releaser ([here](https://github.com/goreleaser/goreleaser/blob/2bdfbdcbfe015c8175ece14efccaf2c610bc2daa/internal/git/git.go) is the internal package inside that project). I _think_ it's a safe assumption that `git` will be a present in nearly all environments, but as you said this change supports the minority use cases as well.

matglas:
I had been thinking about that indeed but had not added it yet. It might be valuable to add the following indeed.

```json
{
  "git-tool": "go-git+go-bin",
  "git-bin-path": "/usr/local/bin",
  "git-bin-hash": "sha256:0abc123..."
```

matglas:
Besides the earlier mentioned points. Does anyone think this needs any specific tests or documentation. I could not really think of anything.

matglas:
I added the output. The data looks like this now. I'm following the naming convention that is used by the other fields.

```json
{
         ...
         "gittool": "go-git+git-bin",
         "gitbinpath": "/opt/homebrew/bin/git",
         "gitbinhash": "sha256:e4c3a12a8af0e9cced8b3a2993d18a5c8ac994cfd4713e7d73d8d2c3c1ea3eba",
         ...
}
```

I am in doubt about the gitbinhash. If I should use the digestset output. But it feels like a lot of overhead to add. I could be wrong and the structure would be better because then it becomes easier to write validators. Let me know your opinions on that.

attestation/git/git_bin.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2023 The Witness Contributors
```
kairoaraujo, line 1:
```suggestion
// Copyright 2024 The Witness Contributors
```

kairoaraujo (dismissed):
Just a nit thing

kairoaraujo (dismissed):
LGTM, I will wait for others review.

colek42:
> I added the output. The data looks like this now. I'm following the naming convention that is used by the other fields.
> 
> ```json
> {
>          ...
>          "gittool": "go-git+git-bin",
>          "gitbinpath": "/opt/homebrew/bin/git",
>          "gitbinhash": "sha256:e4c3a12a8af0e9cced8b3a2993d18a5c8ac994cfd4713e7d73d8d2c3c1ea3eba",
>          ...
> }
> ```
> 
> I am in doubt about the gitbinhash. If I should use the digestset output. But it feels like a lot of overhead to add. I could be wrong and the structure would be better because then it becomes easier to write validators. Let me know your opinions on that.

I think digest set is the way to go, mainly since it supports multiple algos and is pluggable.  cc @mikhailswift

matglas:
Output looks like this now.

```json
{
          "gittool": "go-git+git-bin",
          "gitbinpath": "/opt/homebrew/bin/git",
          "gitbinhash": {
            "sha256": "e4c3a12a8af0e9cced8b3a2993d18a5c8ac994cfd4713e7d73d8d2c3c1ea3eba"
          },
}
```

mikhailswift (dismissed):
Thanks for this, Matias!

mikhailswift:
Just saw that the verify schema CI job failed -- a quick `make schema` should fix it :)

matglas:
Will do that tomorrow. Something went wrong in my local rebase.

mikhailswift (approved):

PR #361, opened 2024-09-26
feat: adding git remotes to subjects and normalising url

ChaosInTheCRD:

Description

This PR adds the repository remotes to the subjects of the in-toto statement for more easy querying of understanding the identity of a repository. It also normalises the remote url, in an attempt to make it agnostic to the protocol used in that config to connect to the remote (e.g., `http://` / `ssh`).

### DIFF
--- a/attestation/git/git_test.go
+++ b/attestation/git/git_test.go
@@ -43,6 +43,25 @@ func TestNameTypeRunType(t *testing.T) {
 	require.Equal(t, RunType, attestor.RunType(), "Expected the attestor's run type")
 }
 
+func TestNormaliseRemotes(t *testing.T) {
+	remotes := []string{
+		"github.com/in-toto/go-witness",
+		"github.com/in-toto/go-witness.git",
+		"https://github.com/in-toto/go-witness",
+		"https://github.com/in-toto/go-witness.git",
+		"git@github.com:in-toto/go-witness",
+		"git@github.com:in-toto/go-witness.git",
+	}
+
+	for _, r := range remotes {
+		nr, err := normaliseRemoteURL(r)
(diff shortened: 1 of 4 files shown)


attestation/git/git_test.go
```diff
@@ -43,6 +43,25 @@ func TestNameTypeRunType(t *testing.T) {
 	require.Equal(t, RunType, attestor.RunType(), "Expected the attestor's run type")
 }
 
+func TestNormaliseRemotes(t *testing.T) {
+	remotes := []string{
+		"github.com/in-toto/go-witness",
+		"github.com/in-toto/go-witness.git",
+		"https://github.com/in-toto/go-witness",
+		"https://github.com/in-toto/go-witness.git",
```
adityasaky, line 51: I think it may be worth testing with `git://` and `git+ssh` etc, even though some of those are deprecated. I think some SCPs (possibly sourceforge?) sets git:// to be default.

go.mod
```diff
@@ -103,6 +103,7 @@ require (
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
 	github.com/tchap/go-patricia/v2 v2.3.1 // indirect
 	github.com/titanous/rocacheck v0.0.0-20171023193734-afe73141d399 // indirect
+	github.com/whilp/git-urls v1.0.0 // indirect
```
adityasaky, line 106: weird this is marked indirect? :thinking: `go mod tidy`?

attestation/git/git_test.go
ChaosInTheCRD: a-ha - I will!

PR #420, opened 2025-01-26
K8s manifest attestor

colek42:

This PR **introduces a brand-new Kubernetes manifest attestor** (`k8smanifest`) to Witness. It enables Witness users to attest Kubernetes manifests (YAML or JSON) by:

1. **Collecting and Splitting** multi-document YAML or JSON arrays into separate documents.  
2. **Applying (optional) Server-Side Dry-Run** via `kubectl` to capture defaulted fields the server would inject.  
3. **Removing Ephemeral Fields** (`uid`, `status`, etc.) and ephemeral annotations to ensure reproducible digests.  
4. **Canonicalizing** the remaining content into stable JSON with `jsoncanonicalizer.Transform`, and  
5. **Computing Digests** to serve as attestation subjects.

**Key Features & Benefits**  
- **Multi-Doc Handling**: Supports multiple Kubernetes objects in a single file (e.g., multi-doc YAML or JSON arrays).  
- **Ephemeral-Field Removal**: Strips fields such as `metadata.uid`, `metadata.resourceVersion`, and `status`—ensuring that transient data doesn't alter the final digests.  
- **Optional Dry-Run**: If `--server-side-dry-run` is used, `kubectl apply --dry-run=server` normalizes resources with server defaults.  
- **Flexible Ignoring**: Allows users to specify additional fields and annotations to ignore.  
- **Thorough Testing**: Includes tests for both YAML and JSON documents, multi-doc scenarios, ephemeral fields, and server-side dry-run.

*(Optional: if there's a related issue, link it here using `fixes #<issue_number>` so that merging this PR closes it.)*

## Acceptance Criteria Met

- [x] Docs changes if needed (usage, CLI flags, etc.)
- [x] Testing changes if needed (new test coverage for multi-doc and ephemeral removal)
- [x] [[DCO Sign-off](https://github.com/apps/dco)](https://github.com/apps/dco)

- Since this attestor did not exist prior to this PR, all code in `k8smanifest/` is net-new.
- The design closely parallels the “structureddata” attestor, to provide a consistent experience in how Witness handles JSON/YAML data plus canonicalization.
- Please review the tests to see how multiple objects in a single file are handled, ephemeral fields are removed, and server-side defaults can be captured.


ChaosInTheCRD:
mentioned this async, but I'd still be pretty keen to integrate some of the container image reference finding capability from https://github.com/chaosinthecrd/dexter into this attestor. Would be pretty nice to know which images were involved in said yaml, add them to the subjects etc.

colek42:
> mentioned this async, but I'd still be pretty keen to integrate some of the container image reference finding capability from https://github.com/chaosinthecrd/dexter into this attestor. Would be pretty nice to know which images were involved in said yaml, add them to the subjects etc.

That is a great idea.

internal/jsoncanonicalizer/es6numfmt.go
```diff
@@ -0,0 +1,71 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+// This package converts numbers in IEEE-754 double precision into the
+// format specified for JSON in EcmaScript Version 6 and forward.
+// The core application for this is canonicalization:
+// https://tools.ietf.org/html/draft-rundgren-json-canonicalization-scheme-02
+
+package jsoncanonicalizer
```
mikhailswift, line 22: I'm a little hesitant about copy/pasting all of this code into the go-witness codebase. I'm also curious if we even need to canonicalize the manifest at all

attestation/k8smanifest/k8s.go
```diff
@@ -0,0 +1,720 @@
...
+}
+
+// Recorded image stores the details of images found in kubernetes manifests
+type RecordedImage struct {
+	Reference string            `json:"reference"`
+	Digest    map[string]string `json:"digest"`
+}
+
+// Attestor implements the Witness Attestor interface for Kubernetes manifests.
+type Attestor struct {
```
mikhailswift, line 95: We should change all of the jsontags to be consistent with the rest of our attestors. Just in this one struct we have snake case, upper camel case, and lower camel case. In other attestors we tend to prefer `ignorefields` and `ignoreannotations`, but I think `ignoreFields` and `ignoreAnnotations` is also acceptable.

mikhailswift (commented):
I have to test and run this yet. 

It also seems like this assumes you have a sane kubectl installed to execute some things.

internal/jsoncanonicalizer/es6numfmt.go
ChaosInTheCRD: So we generate a digest of all the manifests captured in the step. I suppose the question is, do we need it? My answer here is, "we _could_ have it, but it doesn't really mean anything". Are we going to use when querying / verifying? Probably not.

attestation/k8smanifest/k8s.go
ChaosInTheCRD: Sorry, didn't notice this. I will fix.

ChaosInTheCRD:
CI is failing because we are relying on opa <v1, which is deprecated. We should explore updating everything to rego v1 but it should not be in this PR.

ChaosInTheCRD (approved):

ChaosInTheCRD (dismissed):

PR #421, opened 2025-01-27
feat: Implement system packages attestation for Debian and RPM backends

fkautz:
Signed-off-by: Frederick F. Kautz IV <frederick@testifysec.com>


fkautz:
Will add a subject then we will be good for review and merge.

attestation/system-packages/system-packages.go
```diff
@@ -0,0 +1,161 @@
...
+		return "", "", "", err
+	}
+
+	return "linux", distribution, version, nil
+}
+
+// Attest implements attestation.Attestor.
+func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
+	os, dist, version, err := a.backend.DetermineOSInfo()
+	fmt.Println(os)
```
colek42, line 124: Can we use the logger here?  I have a console logger I am using to test this branch.  https://github.com/in-toto/go-witness/pull/408

colek42 (commented):
This looks pretty good.  I'd like to see some table tests to test the parsing functions, check for edge cases, and use of the logger rather than fmt.

colek42 (approved):

PR #464, opened 2025-03-26
ci: add witness build integration and workflow improvements

colek42:
## Summary
- Adds makefile targets for building witness CLI with local go-witness code
- Integrates witness build targets into main Makefile for easier discovery
- Adds workflow for testing the build process
- Displays available attestors after build for verification

## Test plan
- Build witness CLI using the new makefile targets
- Verify the workflow runs successfully

### DIFF
--- a/.github/workflows/build-witness-test.yml
+++ b/.github/workflows/build-witness-test.yml
@@ -1,0 +1,29 @@
+name: Build Witness CLI Test
+
+on:
+  pull_request:
+    branches: [ main ]
+  push:
+    branches: [ main ]
+
+jobs:
+  build-witness:
+    name: Build Witness CLI with local code
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "stable"
+
+      - name: Build Witness CLI with local code
+        run: |
+          # In CI we'll use the checked out copy of go-witness
+          make build-witness GO_WITNESS_PATH=$GITHUB_WORKSPACE
+          
+          # List available attestors
+          echo "Available attestors:"
+          /tmp/witness-build-test/witness attestors list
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -27,6 +27,46 @@ test: ## Run the go unit tests
...
+	@echo "Binary located at: $(WITNESS_TMP_DIR)/witness"
+	
+	# Verify the binary works
+	$(WITNESS_TMP_DIR)/witness version
+	@echo "Available attestors:"
+	$(WITNESS_TMP_DIR)/witness attestors list
+
+.PHONY: clean-witness
+clean-witness: ## Clean up witness build environment
+	rm -rf $(WITNESS_TMP_DIR)
```
matglas, line 68: Should this also revert changes to go.mod which where edited?

.github/workflows/build-witness-test.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    branches: [ main ]
+
+jobs:
+  build-witness:
+    name: Build Witness CLI with local code
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
matglas, line 15: Could we use pinned actions?

.github/workflows/build-witness-test.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
```
matglas, line 18: Could we use pinned actions?

.github/workflows/build-witness-test.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          go-version: "stable"
```
matglas, line 20: We are using a pinned version on witness build. That might be better here too: https://github.com/in-toto/go-witness/blob/762a5c5f431b8a5fba326c6bd3236c82c70a70ff/.github/workflows/witness.yml#L59

matglas (commented):
I like the approach especially building with a similar branch. I only wonder what happens if that branch is not there. Also I see a few things that could be improved.

PR #507, opened 2025-05-23
feat: add MultiExporter interface for multiple attestation exports

colek42:
## Summary

This PR introduces a new `MultiExporter` interface that allows attestors to export multiple individual attestations instead of being limited to a single attestation.

## Motivation

Some attestors need to create separate attestations for each item they process. For example, a product archive attestor might want to create one attestation per output file rather than bundling all files into a single large attestation.

### New Interface
- Added `MultiExporter` interface with `Export()` and `ExportedAttestations()` methods
- Added `ExportedAttestation` struct to represent individual exported attestations
- Each exported attestation includes:
  - Predicate (the attestation data)
  - PredicateType (the attestation type URI)
  - Subjects (for indexing)
  - Name (optional identifier)

### Updated Export Logic
- Modified `run.go` to check for `MultiExporter` first, then fall back to `Exporter`
- Creates individual signed envelopes for each exported attestation
- Attestor names include the export name (e.g., `attestor-name/export-item`)

## Benefits

- Enables fine-grained attestations
- Allows independent verification of individual items
- Improves scalability by avoiding large monolithic attestations
- Maintains backward compatibility with existing `Exporter` interface

## Example Usage

```go
type MyMultiExporter struct{}

func (m *MyMultiExporter) Export() bool {
    return true
}

func (m *MyMultiExporter) ExportedAttestations() []attestation.ExportedAttestation {
    return []attestation.ExportedAttestation{
        {
            Predicate: myData1,
            PredicateType: "https://example.com/predicate/v1",
            Subjects: subjects1,
            Name: "item1",
        },
        {
            Predicate: myData2,
            PredicateType: "https://example.com/predicate/v1",
            Subjects: subjects2,
            Name: "item2",
        },
    }
}
```

## Testing

The existing export tests continue to pass. Additional testing will be included with attestors that implement this interface.


colek42:
@mikhailswift Thanks for the feedback\! I've refactored the implementation based on your suggestions:

- Removed the `ExportedAttestation` struct entirely
- Changed `MultiExporter` interface to return `[]Attestor` instead
- Updated `run.go` to work with attestor instances directly

This approach is much cleaner as it reuses the existing `Attestor` interface. Each exported attestor:
- Has its own `Type()` for the predicate type
- Can implement `Subjecter` for providing subjects
- Has a `Name()` for identification
- Serves as the predicate itself when marshaled

The implementation now leverages existing abstractions without introducing redundant structures.

run.go
```diff
@@ -140,29 +140,72 @@ func run(stepName string, opts []RunOption) ([]RunResult, error) {
...
 			}
 		}
 	}
-
 	if !ro.ignoreErrors && len(errs) > 0 {
 		errs := append([]error{errors.New("attestors failed with error messages")}, errs...)
 		return result, errors.Join(errs...)
 	}
 
+	// Filter attestors for collection - exclude those that implement Exporter and return false
```
mikhailswift, line 190: is this logic right? if it's not an export wouldn't we want it to be part of the collection?

mikhailswift (approved):

PR #529, opened 2025-07-10
feat: Add Azure Key Vault KMS core implementation

fkautz:
## Description

This PR adds core Azure Key Vault KMS support to go-witness. This is the first of several PRs to add complete Azure KMS functionality.

## What's included

- Azure Key Vault client with authentication using DefaultAzureCredential
- SignerVerifier interface implementation for Azure Key Vault
- Full support for RSA (2048/3072/4096) keys
- URI parsing for `azurekms://` scheme
- Basic tests for URI parsing and validation

## Testing

- [x] Unit tests pass
- [x] `go vet` clean
- [x] `go fmt` applied
- [x] Race detector clean

## Known limitations

ECDSA support is partially implemented but signature format conversion will be added in a follow-up PR. The code correctly identifies ECDSA keys but will return an error if used for signing.

## Related PRs

This is part of a series to implement Azure Key Vault support:
- [ ] PR 1: Core implementation (this PR)
- [ ] PR 2: ECDSA signature format conversion
- [ ] PR 3: Comprehensive unit tests
- [ ] PR 4: Integration tests
- [ ] PR 5: Documentation


mikhailswift (approved):
Haven't run or tested this, just did a look over the code. Approving this draft to indicate the direction looks good!

signer/kms/azure/signer.go
```diff
@@ -0,0 +1,142 @@
...
+func (a *SignerVerifier) KeyID() (string, error) {
+	return a.reference, nil
+}
+
+// Sign signs the provided message using Azure Key Vault. If the message is provided,
+// this method will compute the digest according to the hash function specified
+// when the Signer was created.
+func (a *SignerVerifier) Sign(message io.Reader) ([]byte, error) {
+	var err error
+	ctx := context.TODO()
```
jkjell, line 91: We've used `context.Background()` in the GCP KMS implementation. https://github.com/in-toto/go-witness/blob/main/signer/kms/gcp/signer.go#L79

Generally speaking, we've not been very good or consistent in how to allow passing context through to signer/verifiers. @mikhailswift did you have any thoughts on this?

jkjell (commented):
Echoing @mikhailswift's sentiments, this is exactly the right direction. I don't have any comments specific to this implementation just more context sharing.

One consistent issue (feedback areas) with the AWS/GCP KMS was the choice to store the KMS ref by default instead of the public key. This requires read access to the KMS for later verification. I don't have strong opinions on this, other than the fact it's really hard to "undo" the decision later. 😅 

Also, with respect to the proposed PR structure, I'm ok having more complete unit and integration testing in follow on PRs but, it would be great to have a single happy path integration test in the first PR.

signer/kms/azure/signer.go
mikhailswift: We definitely need to do a full pass on adding ctxs as parameters of the functions instead of grabbing todo/background. It’ll probably be a bit painful, affect multiple interfaces, and be a large changeset, though. But I think it’s likely the right way to get contexts into the functions that need them

matglas: Maybe an inexperienced question but does this mean introducing a Context at the 'root' level and passing it along to all the places that could use it? And could this be 'refactored' piece by piece, so the change sets stay manageable?

mikhailswift: The biggest pain in adding context around would be adding it to interfaces that don't currently have it -- since we'd have to ensure all the current implementations of that interface get updated appropriately. Not a difficult task, just one that could somewhat spiral.

Otherwise yeah, you've got the right of it. For the most part in go-witness we don't need to create a context to pass in, users of the library should be providing it, and we just have to fish it down where it needs to be.

colek42 (approved):

PR #539, opened 2025-07-30
fix: Use StaticTokenGetter for pre-fetched tokens to avoid 2-minute timeout

colek42:
## Description

This PR fixes a critical issue where witness times out after exactly 2 minutes when used in GitHub Actions or other automated environments with long-running commands.

## Problem

When witness has a token available (from GitHub Actions OIDC, a file, or command-line flag), it correctly fetches the token but then never uses it with `StaticTokenGetter`. This causes commands to timeout after 2 minutes due to the hardcoded timeout in the interactive OAuth flow.

## Root Cause

The interactive OAuth flow has a 120-second timeout in the sigstore library. When witness fetches a token but doesn't use StaticTokenGetter, it can fall back to the interactive flow which times out.

## Solution

This PR modifies the Fulcio signer to use `StaticTokenGetter` when a token is already available, bypassing the interactive flow entirely.

## Changes

1. **Added logic to use StaticTokenGetter**: When a token is available and we're not in an interactive terminal, use `StaticTokenGetter` instead of the interactive flow
2. **Added tests**: 
   - `TestGitHubActionsTokenTimeout` - Demonstrates the timeout issue
   - `TestStaticTokenUsage` - Verifies static tokens are handled correctly
3. **Added CI workflow**: `.github/workflows/test-fulcio-timeout.yml` to test the fix in real GitHub Actions environment

## Testing

The new CI workflow demonstrates:
- The bug exists in current code (timeout after 2 minutes)
- The fix resolves the issue
- Real GitHub Actions OIDC tokens work correctly

## Impact

This fix is critical for:
- GitHub Actions workflows using witness-run-action
- Any automated environment where commands take >2 minutes
- CI/CD pipelines with long-running test suites or builds

Fixes #627

## Verification

You can verify the issue by running any command that takes more than 2 minutes with witness in GitHub Actions. With current code it will timeout after exactly 2 minutes. With this fix, long-running commands complete successfully.


colek42:
Closing this PR. After further investigation, the OAuth interactive flow shouldn't be triggered in GitHub Actions environments (non-TTY). The 2-minute timeout issue likely has a different root cause.

PR #565, opened 2025-09-18
log failed attestor name for better debugging

rabajaj0509:
Instead of generic error messages, each attestor error now includes the specific attestor name that failed

Description
When the witness CLI logs log.Debugf("attestor execution failed: %v", err), it will now show messages like "attestor git failed: repository does not exist" instead of just "repository does not exist"


colek42 (approved):
lgtm

PR #584, opened 2025-10-17
feat: add configuration attestor

### LINKED ISSUE #414: [Feat]: Configuration Attestor
**Describe the solution you'd like:**

A pre-material attestor could capture the configuration of the `witness run` operation to ensure the command was observed with the proper set of configuration values.

**User value:**

This will allow an additional layer of protections to subsequent policy verification to ensure the intent of process observation matches reality.

**Expected behavior:**

A `configuration` attestor captures relevant CLI parameters and stores them in the `attestation-collection`

**Testing changes required:**

Unit tests of the new attestor and policy test to ensure functionality

**Documentation changes required:**

Document json schema and context of the new attestor.

manzil-infinity180:

Description
A configuration attestor captures the configuration that governed a witness run: the raw command line, the fully-resolved configuration (flags + config file merged), the config file itself (path, digest, content), and the effective post-parse configuration of each attestor.

Updated Examples -
Running on Mac -

```yaml
// test.yaml

swf (main) $ cat test.yaml
run:
    trace: false
    attestor-product-exclude-glob: "*.att.json"
```

```ts
$ witness-dev run --step build -c test.yaml --signer-file-key-path testkey.pem -o build.att.json -a configuration --attestor-product-exclude-glob "*.log" -- go build -o testapp .

{
  "type": "https://witness.dev/attestations/configuration/v0.1",
  "attestation": {
    "command_line": [
      "witness-dev", "run", "--step", "build", "-c", "test.yaml",
      "--signer-file-key-path", "testkey.pem", "-o", "build.att.json",
      "-a", "configuration", "--attestor-product-exclude-glob", "*.log",
      "--", "go", "build", "-o", "testapp", "."
    ],
    "resolved": {
      "step":                          { "value": "build", "source": "commandline" },
      "attestations":                  { "value": ["configuration"], "source": "commandline" },
      "signer-file-key-path":          { "value": "testkey.pem", "source": "commandline" },
      "attestor-product-exclude-glob": { "value": "*.log", "source": "commandline" },
      "trace":                         { "value": "false", "source": "commandline" },
      "attestor-product-include-glob": { "value": "*", "source": "default" }
      // ... every flag of the run command, each tagged with its source
    },
    "config_path": "test.yaml",
    "config_digest": {
      "sha256": "6a884b5955ecf35d8f03f79be55d4247d33730711b96fef0470ec2e17dd2f05f"
    },
    "config_content": {
      "run": {
        "attestor-product-exclude-glob": "*.att.json",
        "trace": false
      }
    },
    "attestors": {
      "product": {
        "exclude-glob": "*.log",
        "include-glob": "*"
      }
    },
    "working_directory": "/Users/rahulxf/JourneyToXYZ/hello-world-witness"
  },
  "starttime": "2026-08-07T00:26:51.784235+05:30",
  "endtime": "2026-08-07T00:26:51.78437+05:30"
}
```

Note in the example above: the config file set exclude-glob: "*.att.json" but the CLI overrode it with "*.log". config_content records what the file claimed while attestors.product records what actually ran, so a policy can detect when a build was not governed by the config file it carried.

Library usage (no CLI) works without any injection. command_line and resolved stay empty and attestor configs are still collected via the Configurer interface.

Fixes #414

### todo:
- [x] Unit tests of the new attestor
- [ ] Follow-up PR in in-toto/witness to inject CLI config (small change in cmd/run.go, tested locally end to end)
- [ ] More attestors implementing Configurer (incremental)

## Acceptance Criteria Met

- [x] Docs changes if needed

### DIFF
--- a/attestation/configuration/configuration.go
+++ b/attestation/configuration/configuration.go
@@ -15,13 +15,14 @@
 package configuration
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
+	"crypto"
 	"os"
 	"strings"
 
 	"github.com/in-toto/go-witness/attestation"
+	"github.com/in-toto/go-witness/cryptoutil"
 	"github.com/invopop/jsonschema"
+	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -51,10 +52,11 @@ func init() {
 }
 
 type Attestor struct {
-	Flags        map[string]string `json:"flags,omitempty"`
-	ConfigPath   string            `json:"config_path,omitempty"`
-	ConfigDigest string            `json:"config_digest,omitempty"`
-	WorkingDir   string            `json:"working_directory,omitempty"`
+	Flags         map[string]string      `json:"flags,omitempty"`
+	ConfigPath    string                 `json:"config_path,omitempty"`
+	ConfigDigest  cryptoutil.DigestSet   `json:"config_digest,omitempty"`
+	ConfigContent map[string]interface{} `json:"config_content,omitempty"`
+	WorkingDir    string                 `json:"working_directory,omitempty"`
 
 	osArgs func() []string
 }
@@ -116,10 +118,20 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
 		a.ConfigPath = ".witness.yaml"
 	}
 
-	// Config digest if file exists
+	// Config digest and content if file exists
 	if data, err := os.ReadFile(a.ConfigPath); err == nil {
-		sum := sha256.Sum256(data)
-		a.ConfigDigest = hex.EncodeToString(sum[:])
+		digestSet, err := cryptoutil.CalculateDigestSetFromBytes(data, []cryptoutil.DigestValue{
+			{Hash: crypto.SHA256, GitOID: false, DirHash: false},
+		})
+		if err == nil {
+			a.ConfigDigest = digestSet
+		}
+
+		// Parse and store config content
+		var configData map[string]interface{}
+		if err := yaml.Unmarshal(data, &configData); err == nil {
+			a.ConfigContent = configData
+		}
 	}
 
 	return nil
--- a/attestation/configuration/configuration.go
+++ b/attestation/configuration/configuration.go
@@ -1,0 +1,184 @@
+// Copyright 2025 The Witness Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configuration
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"os"
+	"strings"
+
+	"github.com/in-toto/go-witness/attestation"
+	"github.com/invopop/jsonschema"
+)
+
+const (
+	Name    = "configuration"
+	Type    = "https://witness.dev/attestations/configuration/v0.1"
+	RunType = attestation.PreMaterialRunType
+)
+
+// This is a hacky way to create a compile time error in case the attestor
+// doesn't implement the expected interfaces.
+var (
+	_ attestation.Attestor  = &Attestor{}
+	_ ConfigurationAttestor = &Attestor{}
+)
+
+type ConfigurationAttestor interface {
+	// Attestor
+	Name() string
+	Type() string
+	RunType() attestation.RunType
+	Attest(ctx *attestation.AttestationContext) error
+	Data() *Attestor
+}
+
+func init() {
+	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor { return New() })
+}
+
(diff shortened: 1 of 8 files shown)


colek42:
When we add this to Witness, let's add it as an option for the next minor version.   I think we should have this be a default attestor once we battle test it a bit.

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func init() {
+	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor { return New() })
+}
+
+type Attestor struct {
+	Flags        map[string]string `json:"flags,omitempty"`
+	ConfigPath   string            `json:"config_path,omitempty"`
+	ConfigDigest string            `json:"config_digest,omitempty"`
```
colek42: This should be a digestset

manzil-infinity180:
It's now ready to review - please review

cc: @colek42 @jkjell @matglas

colek42 (changes requested):

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,196 @@
...
+	} else if v, ok := a.Flags["c"]; ok && v != "" {
+		a.ConfigPath = v
+	} else {
+		a.ConfigPath = ".witness.yaml"
+	}
+
+	// Config digest and content if file exists
+	if data, err := os.ReadFile(a.ConfigPath); err == nil {
+		digestSet, err := cryptoutil.CalculateDigestSetFromBytes(data, []cryptoutil.DigestValue{
+			{Hash: crypto.SHA256, GitOID: false, DirHash: false},
```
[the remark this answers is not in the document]
manzil-infinity180: Fixed - any other changes required?

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,193 @@
...
+		a.WorkingDir = wd
+	}
+
+	// Config path: flag or default
+	if v, ok := a.Flags["config"]; ok && v != "" {
+		a.ConfigPath = v
+	} else if v, ok := a.Flags["c"]; ok && v != "" {
+		a.ConfigPath = v
+	} else {
+		a.ConfigPath = ".witness.yaml"
```
mikhailswift: I think we may need to think of a way to differentiate between the config being loaded from the default file or no file being used at all. However, I think we may have removed the capability of having a default file due to it being a vulernability. @jkjell can you confirm?

jkjell: Yeah, we just removed the default loading of `.witness.yaml` in the most recent release.

manzil-infinity180: done

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,193 @@
...
+	return jsonschema.Reflect(&a)
+}
+
+func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
+	args := a.osArgs()
+	witnessArgs := extractWitnessArgs(args)
+	a.Flags = parseFlags(witnessArgs)
+
+	// Capture working directory
+	if wd, err := os.Getwd(); err == nil {
```
jkjell: I think this should be ctx.WorkingDir() if it's not an empty string. If it is, then we should capture os.Getwd()

jkjell (commented):
This is looking pretty close!

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,195 @@
...
+func (a *Attestor) RunType() attestation.RunType {
+	return RunType
+}
+
+func (a *Attestor) Schema() *jsonschema.Schema {
+	return jsonschema.Reflect(&a)
+}
+
+func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
+	args := a.osArgs()
```
jkjell: This makes an assumption that the configuration is passed in via arguments to a CLI (likely `witness`). This won't work for go-witness being used directly and attestors configured via the library. Also, `witness` uses [viper](https://github.com/spf13/viper) which can allow binding to environment variables instead of CLI arguments.

attestation/configuration/configuration.go
```diff
@@ -0,0 +1,195 @@
...
+		if arg == "--" {
+			return args[:i]
+		}
+	}
+	return args
+}
+
+// parseFlags parses command line flags into a map
+func parseFlags(cmd []string) map[string]string {
+	flags := make(map[string]string)
```
jkjell: Using a `map[string]string` is going to end up overwriting duplicate flags. The most common/frequent example of this is the `-a` flag for adding attestors. A command like `witness run -a github -a git -a slsa` would be an example.

jkjell (commented):
I'll have to apologize for this feedback coming so late after I've looked at this PR a few times. After seeing the issue with `workDir` config, I went through in a bit deeper detail, looking for similar things. 

Based on the feedback from this review, I'm not sure the best path forward. I haven't thought deeply about either of these options but, the paths forward that stick out in my mind are:

1. Pass in the viper (or other style config) to the config attestor - before being parsed by the attestor opts functions
2. Have each attestor record their config as a part of their schema
3. Have a function for each attestor to output their config to be collected by configuration attestor.

@mikhailswift do you have any other thoughts on this?

colek42:
I like option 3 -- it is the cleanest; however, we miss the global config.  Prob needs to be option 1, maybe with some of option 2/3.  Thanks for the review @jkjell

manzil-infinity180:
Thanks @jkjell @colek42 for the review. I reworked the attestor based on this feedback, going with option 1 combined with option 3.

Also, I updated the PR description with an updated example

PR #627, opened 2025-12-25
Add `StagePreExec`, `StagePreExit` hooks for `ExecuteRunType` attestor

Vyom-Yadav:

Description

This is prefaced with - `commandrun` is a special attestor (default as well) that executes the actual command. 

`networktrace` attestor requires to coordinate with the `commanrun` attestor to initiate and end network recording. Specifically, it needs the PID of the command to update some state and before the command exits, it needs to stop the recording to prevent over recording of stuff due to PID recycling (rare but possible). 

There was no mechanism right now, which allowed attestors of same type to coordinate with each other. Rather than creating some very custom for these two attestors, I've created a generic framework in which every attestor declares their intent and registers execute hooks which are executed by the `commandrun` attestor. 

The basic flow is other attestors `Declare` hook followed by actually `Register`ing them.

Since we run all attestors in a different goroutine, declaring first is required to prevent `commandrun` from running the hooks before all hooks are registered. `commandrun` will wait for all hooks to be registered before firing. This API is trying to minimize and mistakes that developers might make while registering hooks and their order of execution. 

Pre-requisite for adding `networktrace` attestor.

## Acceptance Criteria Met

- [ ] Docs changes if needed
- [ ] Testing changes if needed

This is an API design I had in mind, quite open to improving this.

### DIFF
--- a/attestation/context.go
+++ b/attestation/context.go
@@ -146,6 +146,7 @@ type AttestationContext struct {
 	mutex               sync.RWMutex
 	environmentCapturer *environment.Capture
 	outputWriters       []io.Writer
+	executeHooks        ExecuteHooks
 }
 
 type Product struct {
@@ -201,6 +202,25 @@ func (ctx *AttestationContext) RunAttestors() error {
 	for _, k := range order {
 		log.Infof("Starting %s attestors stage...", k.String())
 
+		// Phase 1: For ExecuteRunType, declare hooks before starting goroutines
+		if k == ExecuteRunType {
+			for _, att := range attestors[k] {
+				if declarer, ok := att.(ExecuteHookDeclarer); ok {
+					if err := declarer.DeclareHooks(&ctx.executeHooks); err != nil {
+						return fmt.Errorf("attestor %s failed to declare hooks: %w", att.Name(), err)
+					}
+				}
+			}
+		} else {
+			// Verify non-ExecuteRunType attestors don't implement ExecuteHookDeclarer
+			for _, att := range attestors[k] {
+				if _, ok := att.(ExecuteHookDeclarer); ok {
+					return fmt.Errorf("attestor %s of run type %s cannot declare execute hooks", att.Name(), k.String())
+				}
+			}
+		}
+
+		// Phase 2: Run attestors concurrently
 		var wg sync.WaitGroup
 
 		for _, att := range attestors[k] {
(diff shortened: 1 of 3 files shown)


Vyom-Yadav:
cc @jkjell @mikhailswift

attestation/factory.go
```diff
@@ -83,6 +83,12 @@ type BackReffer interface {
 	BackRefs() map[string]cryptoutil.DigestSet
 }
 
+// ExecuteHookDeclarer allows attestors to declare execute hooks. DeclareHooks is called
+// before running the attestor that implements this interface.
+type ExecuteHookDeclarer interface {
+	DeclareHooks(hooks *ExecuteHooks) error
+}
```
Vyom-Yadav: Hmm, for `commandrun` to access the hooks, it'd need to implement this interface (as it's not exposed via `ctx *attestation.AttestationContext`). All clean code practices say me to declare another interface - `ExecuteHookRunner` which would be only implemented by `commandrun` to gain access to the hooks, however, the same can be achieved by implementing this interface (as a special case) which allows us to save creation of another interface and:

```go
if err := declarer.DeclareHooks(&ctx.executeHooks); err != nil {
  return fmt.Errorf("attestor %s failed to declare hooks: %w", att.Name(), err)
}
```

does the job for both. Open to suggestions.

attestation/context.go
```diff
@@ -201,6 +202,25 @@ func (ctx *AttestationContext) RunAttestors() error {
...
+					if err := declarer.DeclareHooks(&ctx.executeHooks); err != nil {
+						return fmt.Errorf("attestor %s failed to declare hooks: %w", att.Name(), err)
+					}
+				}
+			}
+		} else {
+			// Verify non-ExecuteRunType attestors don't implement ExecuteHookDeclarer
+			for _, att := range attestors[k] {
+				if _, ok := att.(ExecuteHookDeclarer); ok {
+					return fmt.Errorf("attestor %s of run type %s cannot declare execute hooks", att.Name(), k.String())
```
mikhailswift: i wonder if we could implement something like this as a linting rule, so we can check it before runtime. not a suggestion for this, just a thought as i read the code

attestation/factory.go
mikhailswift: if i'm understanding this correctly, i think i have a preference for command run to have a function called `RunHooks` and a separate `ExecuteHookRunner`. otherwise the intent of the code gets a bit muddier.

PR #629, opened 2025-12-28
Add `networktrace` attestor with support for invisible TCP proxy

Vyom-Yadav:

Adds `networktrace` attestor with support for invisible TCP proxy. 

Part of https://github.com/in-toto/witness/issues/36

## Acceptance Criteria Met

- [ ] Docs changes if needed


Vyom-Yadav:
I'll cleanup the license and copyright headers stuff and add some integration tests, but I've reduced this to only TCP proxy for making the review manageable. Only 3rd commit (onward) is the code related to network trace attestor. 

cc @stupendoussuperpowers @jkjell @mikhailswift

Vyom-Yadav:
@stupendoussuperpowers @mikhailswift @jkjell Ready for review whenever you folks have time

attestation/networktrace/bpf/connect.bpf.c
```diff
@@ -0,0 +1,165 @@
...
+    __u64 cookie = bpf_get_socket_cookie(ctx);
+    __u32 pid = bpf_get_current_pid_tgid() >> 32;
+    __u64 cgroup_id = bpf_get_current_cgroup_id();
+
+    char comm[MAX_COMM_LEN];
+    bpf_get_current_comm(&comm, sizeof(comm));
+
+    int should_intercept_result = should_intercept(pid, cgroup_id, comm);
+
+    if (!should_intercept_result) {
```
stupendoussuperpowers: Flipped true/false val.

attestation/networktrace/bpf/headers/helpers.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    struct comm_allowlist_key comm_key = {};
+    __builtin_memcpy(comm_key.comm, comm, MAX_COMM_LEN);
+    __u8* comm_val = bpf_map_lookup_elem(&comm_allowlist, &comm_key);
+    if (comm_val != NULL) {
+        return 1;
+    }
+
+    return 0;
```
stupendoussuperpowers: the true/false values are flipped here. the connect probe runs `if(!should_intercept) { return 1; }`

attestation/networktrace/bpf/bpf.go
```diff
@@ -0,0 +1,536 @@
...
+	if err := rlimit.RemoveMemlock(); err != nil {
+		return fmt.Errorf("remove memlock: %w", err)
+	}
+
+	for _, dir := range []string{
+		PinBasePath,
+		PinBasePath + "/progs",
+		PinBasePath + "/links",
+	} {
+		if err := os.MkdirAll(dir, 0700); err != nil {
```
stupendoussuperpowers: Add a check here to see if /sys/fs/bpf is mounted so that we have a better error message/mounting logic.

stupendoussuperpowers (commented):
I used this docker container for testing: https://gist.github.com/stupendoussuperpowers/feb0a79bd159c73efb822f4ef4777846

In the current state, I am unable to get this to log any network activity. I had to make a few tweaks to get it to run as expected, which I've detailed towards the end.

Before diving into that, I want to address some general issues:

0. `networktrace` is missing from `imports.go`
1. If `/sys/fs/bpf` is not mounted with `-t bpf`, we see a generic `mkdir` error when loading programs, this should either a) mount this path itself, or b) report an error detailing that this failure was due to bpf not being mounted. 

2. When running the network-trace attestor without the `-r` flag (commandrun tracing), the target program doesn't execute properly. It starts a new PID, but the first ptrace signal causes the program to exit prematurely. It doesn't get reported as a failed attestation run either. 

    As a workaround for testing, I either ran this with `-r` enabled or added a universal `injectedSig = 0` here https://github.com/in-toto/go-witness/blob/3d56d793359bdb3c827a592436c2c9d1994d291b/attestation/commandrun/tracing_linux.go#L243-L247 (neither is a robust solution the second needs way more work). 

We need decisions on:

1. Whether or not network-trace should be tied with `-r`
2. If not, adding a workaround to ensure non `-r`'d network-trace handles ptrace signals properly. 
3. The product attestor has special handling for command-run cases where `-r` is enabled. It does not report the generated artifacts which were not read by the build process. I am unsure on the design philosophy behind this but might be worth considering if network-trace should be treated similarly, specially if its tied with the flag.

https://github.com/in-toto/go-witness/blob/3d56d793359bdb3c827a592436c2c9d1994d291b/attestation/file/file.go#L110-L114

--- 

I was able to make this work and record attesations after making the following changes:

1. `should_intercept` and `is_pid_allowed` checks are flipped in places. Sometimes callers expect it to return 1 for true, sometimes 0 for true, etc. Fixed this in all the places 
2. `get_current_pid_tgid()` always returns the PID on the _host_, so if you run this in a docker container or a namespace in general, the PIDs and their checks are going to be mismatched. I changed that with the approach I was using for my kubecon demo:

More info: https://docs.ebpf.io/linux/helper-function/bpf_get_ns_current_pid_tgid/ (only works on kernel >6.10)
Kubecon implementation (for older kernels): https://github.com/stupendoussuperpowers/witprox/blob/d1d1abf63839b6629dac7cc67ba618258fa29a27/internal/bpf/redirect.bpf.c#L54

3. the `attestation Attestation` field is private, so it doesn't show up in the final attestation output. Adding a `func (n *Attestor) MarshalJSON() ([]byte, error)` implementation fixes that.

attestation/networktrace/bpf/headers/helpers.h
```diff
@@ -0,0 +1,127 @@
...
+#ifndef __HELPERS_H__
+#define __HELPERS_H__
+
+#include "vmlinux.h"
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+
+#include "map_defs.h"
+
+#define MAX_PARENT_WALK 64
```
Vyom-Yadav: There are plans to hook into process creation and populate a map of parent child PIDs in the future. 64 level walk can be defeated by a payload to bypass the scanner.

stupendoussuperpowers (approved):
Works fine now without needing any local changes.

Still seeing the issue where this doesn't run properly if I don't supply `-r/--tracing` flag to `command-run`. That should be fixed separately anyway so will self-assign a PR for that.

Vyom-Yadav:
@jkjell @mikhailswift I've marked network trace as experimental, can we merge this PR?

Vyom-Yadav:
ci failure unrelated.

attestation/networktrace/networktrace.go
```diff
@@ -0,0 +1,291 @@
...
+var (
+	_ attestation.Attestor            = &Attestor{}
+	_ attestation.ExecuteHookDeclarer = &Attestor{}
+)
+
+func init() {
+	attestation.RegisterAttestation(Name, Type, RunType, func() attestation.Attestor {
+		return New()
+	})
+}
```
Vyom-Yadav, line 47: I'll add the configuration options alongside the http proxy code in the next PR.

mikhailswift:
breaks builds on macos

```
package command-line-arguments
	imports github.com/in-toto/witness/cmd
	imports github.com/in-toto/go-witness
	imports github.com/in-toto/go-witness/attestation/networktrace: build constraints exclude all Go files in /Users/mswift/Workspaces/witness-test/go-witness/attestation/networktrace
```

the import in imports.go needs to be guarded behind a build constraint

Vyom-Yadav:
> breaks builds on macos
> 
> ```
> package command-line-arguments
> 	imports github.com/in-toto/witness/cmd
> 	imports github.com/in-toto/go-witness
> 	imports github.com/in-toto/go-witness/attestation/networktrace: build constraints exclude all Go files in /Users/mswift/Workspaces/witness-test/go-witness/attestation/networktrace
> ```
> 
> the import in imports.go needs to be guarded behind a build constraint

Done @mikhailswift

mikhailswift (approved):

PR #649, opened 2026-01-29
feat: support offline KMS verification with embedded public keys

### LINKED ISSUE #648: [Bug]: PublicKeyVerifiers fails with ErrKeyIDMismatch when using embedded public key with KMS key ID
**What steps did you take and what happened:**

When attempting to verify attestations signed with AWS KMS keys in an offline/air-gapped environment, the verification fails with `ErrKeyIDMismatch` even when the policy contains an embedded public key.

Steps to reproduce:
1. Create a policy with a KMS key ID (e.g., `awskms:///26059292-db29-4f01-91d9-b43fc584049c`) and embed the corresponding public key in the `key` field
2. Sign an attestation using the KMS key
3. Attempt to verify the attestation without AWS credentials (simulating an air-gapped environment)

The verification fails in `PublicKeyVerifiers()` at [policy/policy.go lines 96-105](https://github.com/in-toto/go-witness/blob/1453b041480a3878a93a324452e6e1bc6899ac02/policy/policy.go#L96-L105):

```go
if verifier == nil {
    verifier, err = cryptoutil.NewVerifierFromReader(bytes.NewReader(key.Key))
    if err != nil {
        return nil, err
    }
}

keyID, err := verifier.KeyID()
if err != nil {
    return nil, err
}

if keyID != key.KeyID {
    return nil, ErrKeyIDMismatch{
        Expected: key.KeyID,
        Actual:   keyID,
    }
}
```

The bug occurs because:
1. When KMS is not accessible, the code correctly falls back to the embedded key (line 91)
2. However, it then computes `keyID` from the verifier (line 98), which returns a hash of the public key
3. It compares this computed `keyID` against `key.KeyID` (line 102), which is the KMS URI
4. These will **never** match: `awskms:///...` ≠ `31a05aa1c7a1eaa51630a0f5e99caf4a10ae981ea8d16a44370bf7267b0bc392`
5. `ErrKeyIDMismatch` is returned even though the embedded key is valid

**What did you expect to happen:**

When a policy contains both a KMS-style key ID AND an embedded public key, the embedded key should be used for verification without requiring KMS access, and the verification should succeed.

This enables important use cases:
- Air-gapped environments where KMS services are not accessible
- Offline verification workflows
- Reducing latency by avoiding KMS API calls
- CI/CD pipelines that need to verify attestations without KMS credentials

**Anything else you would like to add:**

I have a fix ready that:
1. Prioritizes the embedded key when present (avoiding KMS calls entirely)
2. Skips the keyID mismatch check when using an embedded key with a KMS-style key ID
3. Stores verifiers under the policy key ID (KMS URI) so functionary matching continues to work
4. Includes comprehensive unit tests for the new functionality
5. Is fully backward compatible - existing behavior is preserved when no embedded key is provided

**Environment:**

- Go-Witness version: v0.5.0 and later (tested on main branch)
- Architecture: linux/amd64
- Attestors used: git, material, command-run, product, slsa, sarif
- Archivista version: N/A (local verification)

jessesanford:

This PR fixes a bug in `PublicKeyVerifiers()` that prevents offline/air-gapped verification of attestations signed with KMS keys, even when the policy contains an embedded public key.

### The Bug

In the current implementation at [policy/policy.go lines 96-105](https://github.com/in-toto/go-witness/blob/1453b041480a3878a93a324452e6e1bc6899ac02/policy/policy.go#L96-L105):

1. When KMS is not accessible, the code correctly falls back to using the embedded key
2. However, it then computes `keyID` from the verifier, which returns a hash of the public key (e.g., `31a05aa1c7a1eaa51630a0f5e99caf4a10ae981ea8d16a44370bf7267b0bc392`)
3. It compares this against `key.KeyID`, which is the KMS URI (e.g., `awskms:///26059292-db29-4f01-91d9-b43fc584049c`)
4. These will **never** match, so `ErrKeyIDMismatch` is returned even though the embedded key is valid

Additionally, even after fixing the above, **functionary matching fails** because:
- The policy's functionary expects `PublicKeyID` to match the KMS URI
- But the verifier's `KeyID()` method returns the hash of the embedded public key
- This causes `Functionary.Validate()` to fail with "no verifiers matched with allowed functionaries"

### The Fix

This PR introduces two key changes:

1. **Prioritizes embedded keys**: When a policy entry has an embedded key, use it directly without attempting KMS calls, and skip the keyID mismatch check for KMS+embedded key combinations

2. **`keyIDOverrideVerifier` wrapper**: When creating a verifier from an embedded key for a KMS-style key ID, wrap it with `keyIDOverrideVerifier` that overrides the `KeyID()` method to return the KMS URI instead of the public key hash. This ensures functionary matching works correctly.

### Use Cases Enabled

- Air-gapped environments where KMS services are not accessible
- Offline verification workflows
- Reducing latency by avoiding KMS API calls
- CI/CD pipelines that need to verify attestations without KMS credentials

### Backward Compatibility

This change is fully backward compatible:
- If no embedded key is provided, the existing KMS verification path is used
- The keyID mismatch check is only skipped when BOTH a KMS-style key ID AND an embedded key are present
- Regular (non-KMS) keys are not wrapped and retain their original `KeyID()` behavior

Fixes #648

## Acceptance Criteria Met

- [x] Docs changes if needed (updated function documentation)
- [x] Testing changes if needed (added comprehensive unit tests for offline KMS verification)

The key changes are in `policy/policy.go`:
- `keyIDOverrideVerifier` struct: Wraps a verifier to override `KeyID()` for functionary matching
- Lines 69-88: Prioritize embedded key when present and wrap with `keyIDOverrideVerifier` for KMS keys
- Lines 119-140: Skip keyID mismatch check for KMS+embedded key combinations
- Line 143: Store verifier under the policy key ID (not the computed key ID)

New tests in `policy/policy_test.go`:
- `TestKMSOfflineVerification`: Tests the basic offline verification functionality
- `TestKMSOfflineVerificationWithFunctionaries`: Tests that functionary matching still works
- `TestKeyIDOverrideVerifier`: Tests that the verifier's `KeyID()` returns the KMS URI
- `TestKeyIDOverrideVerifierNotAppliedToNonKMS`: Tests that non-KMS keys are not wrapped

A test tool is also included in `cmd/verify-test/` that demonstrates the fix working with real policy and attestation files.

### DIFF
--- a/cmd/verify-test/main.go
+++ b/cmd/verify-test/main.go
@@ -1,0 +1,156 @@
+// verify-test is a simple command-line tool to test offline KMS verification
+// using the embedded public key feature
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/in-toto/go-witness/cryptoutil"
+	"github.com/in-toto/go-witness/dsse"
+	"github.com/in-toto/go-witness/policy"
+	"github.com/in-toto/go-witness/signer"
+
+	// Import AWS KMS provider
+	_ "github.com/in-toto/go-witness/signer/kms/aws"
+)
+
+func main() {
+	if len(os.Args) < 3 {
+		fmt.Fprintf(os.Stderr, "Usage: %s <policy.json> <attestation.json>\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	policyPath := os.Args[1]
+	attestationPath := os.Args[2]
+
+	// Read the policy file
+	policyData, err := os.ReadFile(policyPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to read policy file: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Check if it's a signed policy (DSSE envelope)
+	var policyEnvelope dsse.Envelope
+	if err := json.Unmarshal(policyData, &policyEnvelope); err == nil && policyEnvelope.PayloadType != "" {
+		// It's a DSSE envelope - the Payload is already decoded from base64 by json.Unmarshal
+		// because it's a []byte field
+		policyData = policyEnvelope.Payload
+		fmt.Println("Decoded signed policy envelope")
+	}
+
+	// Parse the policy
+	var pol policy.Policy
+	if err := json.Unmarshal(policyData, &pol); err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to parse policy: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Policy loaded successfully\n")
+	fmt.Printf("  - Steps: %d\n", len(pol.Steps))
+	fmt.Printf("  - Public Keys: %d\n", len(pol.PublicKeys))
+
+	// Print public key info
+	for keyID, pk := range pol.PublicKeys {
+		fmt.Printf("    - Key ID: %s\n", keyID)
+		fmt.Printf("      Has embedded key: %v (%d bytes)\n", len(pk.Key) > 0, len(pk.Key))
+	}
+
+	// Create verifiers from the policy's public keys
+	// This is the key test - it should work WITHOUT AWS credentials
+	fmt.Println("\n=== Testing PublicKeyVerifiers (the fix) ===")
+	fmt.Println("Creating verifiers from policy public keys...")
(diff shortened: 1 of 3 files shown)


mikhailswift:
Otherwise I think this looks good. It's passing tests I'm running

PR #673, opened 2026-03-12
Drop copy of non-freely licensed DCO from CONTRIBUTORS

jas4711:
Hi!  This is part of the Debian packaging review.  The DCO text is not freely licensed, which means we cannot include it in Debian.  While we could avoid shipping the CONTRIBUTORS file completely, I am hoping this small fix is acceptable to you, which would allow shipping go-witness unmodified in Debian.  Deviating from upstream source code is painful from a supply-chain auditing perspective, even if changes are minor.

```
Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.
```

Thanks,
Simon

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,45 +16,7 @@ specification for commit messages.  Please read and familiarize yourself with th
 work to be brought upstream.
 
 We also ask that you sign-off your commits.  By signing off you agree to the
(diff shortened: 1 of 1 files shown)


jkjell:
I have no personal objection to this change. Just going to confirm it's acceptable with some folks from the CNCF. cc @JustinCappos

PR #683, opened 2026-03-18
Add networktrace attestor hack

jkjell:

Based off of https://github.com/in-toto/go-witness/pull/629.

The last commit has the changes I needed to successfully run:

`sudo ./bin/witness run -l debug -a network-trace -r -k ../testkey.pem -o test-att.json --step build -- bash -c 'curl -4 https://www.google.com > google.html'`

and get an attestation. The `curl` command is only using IPv4 because of an unhandled connection error for IPv6.

### DIFF
--- a/attestation/networktrace/networktrace.go
+++ b/attestation/networktrace/networktrace.go
@@ -65,7 +65,7 @@ type Attestation struct {
 // Attestor implements the network trace attestation
 type Attestor struct {
 	config      types.Config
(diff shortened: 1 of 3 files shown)


attestation/networktrace/networktrace.go
jkjell: Changed `attestation` to `Attestation` to allow it to be serialized into the attestation collection.

PR #740, opened 2026-05-29
[command-run] Add support for eBPF based tracing (and non-ptrace based tracers)

stupendoussuperpowers:
### What
Current support for process/openedfiles tracing in the `command-run` attestor is limited to `ptrace` on Linux. This PR aims to 1. Expand support for other tracing backends, and 2. Add support for eBPF `tracepoint/syscall` based tracking of relevant syscalls such as open/exec/exit. 

Usage: `witness run ... --trace --attestor-command-run-trace-backend ebpf`

The default tracing backend is still `ptrace`, so without supplying the `--attestor-command-run-trace-backend` flag the behavior is going to remain unchanged. 

### Why
The current ptrace-based tracking is substantially slower than both a non-traced witness run and an eBPF-traced witness run. With projects such as SBOMit that require monitoring of opened files for a build, these slower times make adoption impossible for any project of significant size. eBPF is therefore needed to ensure we can trace these events without the kinds of overhead we see with ptrace.

Secondly, the current support for tracing is limited to Linux. By adding a more modular approach to how we can define a tracer, adding support for Windows/BSD/MacOS specific tracing mechanisms in the future is going to be more streamlined. 

## Acceptance Criteria Met

- [ ] Docs changes if needed

To test these changes, you'll need to:

- Ensure eBPF is enabled on your machine.
- If you make any changes to the eBPF code, you'll need to run `make generate-commandrun-bpf` to refresh object files.
- Run using `witness run ... --trace --attestor-command-run-trace-backend ebpf`.
- To run integration test, `go test -tags=ebpf_integration ./attestation/commandrun/...`.


attestation/commandrun/bpf/gen.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build linux
+
+package bpf
+
+//go:generate sh -c "go tool bpf2go -cc clang -no-strip -target bpfel,bpfeb -go-package bpf filetraceSyscall filetrace_syscall.bpf.c -- -Wall -Werror -I ../../bpf-common/headers -I \"$(go env GOPATH)/pkg/mod/github.com/cilium/ebpf@$(go list -m -f '{{.Version}}' github.com/cilium/ebpf)/examples/headers\""
```
Vyom-Yadav: Is `-I \"$(go env GOPATH)/pkg/mod/github.com/cilium/ebpf@$(go list -m -f '{{.Version}}' github.com/cilium/ebpf)/examples/headers\""` required? Wouldn't go tool automatically figure that out?

attestation/commandrun/ebpf_linux.go
```diff
@@ -0,0 +1,482 @@
...
+	if len(statusBytes) > 0 {
+		if parsedPPID, err := getPPIDFromStatus(statusBytes); err == nil {
+			ppid = parsedPPID
+		}
+	}
+	comm := ""
+	if len(statusBytes) > 0 {
+		for line := range strings.SplitSeq(string(statusBytes), "\n") {
+			if _, found := strings.CutPrefix(line, "Name:"); found {
+				break
```
Vyom-Yadav: `comm = strings.TrimSpace(strings.TrimPrefix(line, "Name:")) `

attestation/commandrun/bpf/filetrace_common.h
```diff
@@ -0,0 +1,155 @@
...
+	event->error = error;
+	event->path[0] = '\0';
+	bpf_ringbuf_submit(event, 0);
+}
+
+/* sched_process_exec and sched_process_exit give userspace enough lifecycle
+ * information to create/enrich ProcessInfo records around the open events.
+ */
+SEC("tracepoint/sched/sched_process_exec")
+int trace_sched_process_exec(struct sched_process_exec_args *ctx) {
```
Vyom-Yadav: Maybe move the tracepoint hooks to a .c file?

attestation/commandrun/bpf/filetrace_common.h
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	struct file_open_event *event =
+	    bpf_ringbuf_reserve(&events, sizeof(*event), 0);
+	if (!event) {
+		return 0;
+	}
+	event->event_type = EVENT_TYPE_EXIT;
+	event->pid = ctx->pid;
+	event->tid = ctx->pid;
```
Vyom-Yadav: Kernel pid wouldn't be same as tgid or userspace pid, shouldn't this use  `bpf_get_current_pid_tgid`?

attestation/commandrun/bpf/filetrace_common.h
```diff
@@ -0,0 +1,155 @@
...
+struct sched_process_exec_args {
+	__u16 common_type;
+	__u8 common_flags;
+	__u8 common_preempt_count;
+	__s32 common_pid;
+	char filename[256];
+	__s32 pid;
+	__s32 old_pid;
+};
+
```
Vyom-Yadav: This layout doesn't seem correct:

```
 $ sudo cat /sys/kernel/debug/tracing/events/sched/sched_process_exec/format 
name: sched_process_exec
ID: 298
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:__data_loc char[] filename;	offset:8;	size:4;	signed:0;
	field:pid_t pid;	offset:12;	size:4;	signed:1;
	field:pid_t old_pid;	offset:16;	size:4;	signed:1;

print fmt: "filename=%s pid=%d old_pid=%d", __get_str(filename), REC->pid, REC->old_pid
```

Some googling suggests __data_loc you need to read from an offset of lower 16 bits.

attestation/commandrun/bpf/filetrace_common.h
```diff
@@ -0,0 +1,155 @@
...
+	__s32 common_pid;
+	char comm[16];
+	__s32 pid;
+	__s32 prio;
+};
+
+/* Ring buffer that store open, exec, and exit events captured by tracepoints. */
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, 1 << 29);
```
Vyom-Yadav, line 69: 1 << 29 seems a bit high. It's 512 MB. I'm not sure how memory allocation is being done here, but is that really required?

attestation/commandrun/ebpf_linux.go
```diff
@@ -0,0 +1,482 @@
...
+	"github.com/cilium/ebpf/link"
+	"github.com/cilium/ebpf/ringbuf"
+	"github.com/in-toto/go-witness/attestation"
+	"github.com/in-toto/go-witness/cryptoutil"
+	"github.com/in-toto/go-witness/log"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	commandRunTraceCgroupPath = "/sys/fs/cgroup/witness-tracing"
```
Vyom-Yadav, line 42: Although it should be okay for most builds, but this forces the build to execute from within a particular cgroup. A user might not want to do that. Can't think of an alternative right now (late night review) other than tracking the complete process tree, which is tedious.

Vyom-Yadav: Also, maybe making the cgroup path random for every run so multiple witness processes can run parallely.

attestation/commandrun/bpf/filetrace_syscall.bpf.c
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return 0;
+}
+
+// Called on sys_exit for open* syscalls.
+// - If pending_open has an error, trying re-reading the path buffer.
+// - Upon success or failure, dispatch an event to the ring buffer to be decoded
+// outside of eBPF in go.
+static __always_inline int submit_pending_open_event(__s64 ret) {
+	__u64 pid_tgid = bpf_get_current_pid_tgid();
```
Vyom-Yadav: `ret` is not being used. So in every case, `EVENT_TYPE_OPEN` will be submitted to the userspace.

attestation/commandrun/bpf/filetrace_syscall.bpf.c
```diff
@@ -0,0 +1,175 @@
...
+	// Try to read the path buffer as a string on sys_enter for open.
+	// If this fails the pending_open event is marked as such so that it can
+	// be retried at sys_exit.
+	long copied = bpf_probe_read_user_str(pending.path,
+					      sizeof(pending.path), filename);
+	if (copied < 0) {
+		pending.error = copied;
+		pending.path[0] = '\0';
+	} else {
+		pending.error = copied;
```
Vyom-Yadav: Using error field for storing the length looks confusing to me. It works, but maybe it'd be cleaner to store it elsewhere?

attestation/commandrun/ebpf_linux.go
```diff
@@ -0,0 +1,482 @@
...
+	return processes
+}
+
+// Enrich ProcessInfo from /proc. These reads are best
+// effort because exec/exit events can race with process teardown.
+func (p *ebpfTraceContext) populateMetadataForProc(pid int, overwrite bool) {
+	statusBytes, _ := os.ReadFile(fmt.Sprintf("/proc/%d/status", pid))
+	cmdlineBytes, _ := os.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid))
+	exePath, _ := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
+
```
Vyom-Yadav, line 478: Can this be obtained using bpf during the process execution? And instead of storing that as a ring buffer event, add that to a map? And before the process exits, it can try to compute the digest of the exe?

Vyom-Yadav (commented):
Some initial thoughts and few fixes. I'm ok with putting the process under a dedicated cgroup, although not a big fan, but I understand PID tracking is complicated.

attestation/commandrun/commandrun.go
```diff
@@ -197,26 +217,34 @@ func (rc *CommandRun) runCmd(ctx *attestation.AttestationContext) error {
...
+	// Check if the tracer being used is eBPF based.
+	if rc.enableTracing && !needsHookTracing && rc.usesEBPFTracing() {
+		rc.Processes, err = rc.traceWithEBPF(c, ctx)
+	} else {
+		if err := c.Start(); err != nil {
+			return err
+		}
+	}
+
+	if rc.enableTracing && (!rc.usesEBPFTracing() || needsHookTracing) {
```
Vyom-Yadav: I think even with ptrace enabled bpf probes can do the heavy work. ptrace is required to run the hooks, but that portion of ptrace is pretty light. 

Even when ptrace is required, bpf tracing can be done afaik

attestation/commandrun/bpf/filetrace_common.h
stupendoussuperpowers: Yeah the __data_loc read was wrong. Fixed that, but also moved to a non custom rolled struct for exec args.

attestation/commandrun/ebpf_linux.go
stupendoussuperpowers: I think there are limitations to what amount of information can be extracted completely independent of /proc reads. 

For instance I think you can get comm and file path, but their lengths might be truncated, and the paths might not be absolute paths, etc. Calculating digest can/should only happen outside of bpf too. 

I did not change this for now, will look into better fixes if possible and not tedious.

attestation/commandrun/commandrun.go
stupendoussuperpowers: Fixed this so that hooks can be used with this. It only does exit/exec tracing which should be fine for hooks while also making sure we dont see crazy overhead.

attestation/commandrun/ebpf_linux.go
stupendoussuperpowers: Added UUID/random cgroup names for parallel flows.

attestation/commandrun/bpf/filetrace_common.h
stupendoussuperpowers: I need to figure out if there's any heuristics on this. 

There's no real way to know when the ringbuffer was filled faster than it was read and therefore events were overwritten. 1 << 29 is the most generous safety margin I provided here. Agreed that this size is overkill but I dont want to reduce this before I can get some data-based evidence about what a tight margin would be.

Makefile
```diff
@@ -42,18 +42,27 @@ lint: ## Run the linter
...
+	mkdir -p ./attestation/bpf-common/headers && bpftool btf dump file /sys/kernel/btf/vmlinux format c > ./attestation/bpf-common/headers/vmlinux.h
+
+.PHONY: generate-commandrun-bpf
+generate-commandrun-bpf: generate-vmlinux ## Generate BPF bytecode and Go bindings for command-run file tracing
+	@echo "Generating command-run BPF code (requires clang and llvm)..."
+	go generate -tags linux ./attestation/commandrun/bpf/...
 
 .PHONY: generate-bpf
-generate-bpf: generate-vmlinux ## Generate BPF bytecode and Go bindings for network trace attestor
+generate-networktrace-bpf: generate-vmlinux ## Generate BPF bytecode and Go bindings for network trace attestor
```
Vyom-Yadav:
```
.PHONY: generate-networktrace-bpf
generate-networktrace-bpf: generate-vmlinux 
```

Same for the debug one. PHONY name changed, but target hasn't.

attestation/commandrun/bpf/filetrace_syscall.bpf.c
```diff
@@ -0,0 +1,180 @@
+// go:build ignore
+
```
Vyom-Yadav:
```suggestion
//go:build ignore
```

attestation/commandrun/ebpf_linux.go
stupendoussuperpowers: This is not changed to: Using eBPF to generate the initial metadata, and the updating it with /proc whenever possible.

attestation/commandrun/bpf/filetrace_syscall.bpf.c
```diff
@@ -0,0 +1,238 @@
...
+		event->path[0] = '\0';
+	}
+
+	bpf_ringbuf_submit(event, 0);
+	return 0;
+}
+
+SEC("tracepoint/sched/sched_process_exit")
+int trace_sched_process_exit(struct sched_process_exit_args *ctx) {
+	if (!commandrun_in_target_cgroup()) {
```
Vyom-Yadav, line 218: Use `trace_event_raw_sched_process_exit` for uniformity?

stupendoussuperpowers:
Fixed merge issues on `go.mod` and the GHA checks. @jkjell

jkjell (approved):

PR #755, opened 2026-06-09
OCI: detect and parse spec compliant index.json

### LINKED ISSUE #686: [Bug]: OCI Attestor searches for manifest.json instead of index.json breaking OCI Archive format support
**What steps did you take and what happened:**

The OCI attestor in [oci.go](https://github.com/in-toto/go-witness/blob/3d56d793359bdb3c827a592436c2c9d1994d291b/attestation/oci/oci.go#L224) searches for manifest.json inside the tar archive, which is the Docker Archive format. According to the [OCI Image Layout Specification](https://github.com/opencontainers/image-spec/blob/a4c6ade7bb82b316d45391f572727a63e268b252/image-layout.md), the correct entry point for an OCI archive is index.json.

When using buildah push --format oci ... oci-archive:image.tar, the tar contains:

```TEXT
blobs/sha256/...
index.json       ← OCI spec
oci-layout       ← OCI spec
```
This causes the attestor to fail with:

```TEXT
level=debug msg="(attestation/oci) error parsing manifest: unexpected end of JSON input"
level=error msg="attestors failed with error messages\nattestor oci failed: unexpected end of JSON input"
```

**What did you expect to happen:**

The OCI attestor should support index.json (OCI Archive format) in addition to or instead of manifest.json (Docker Archive format).

[Expected outcome listed here.]

**Anything else you would like to add:**

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Go-Witness version: v0.10.2
- Architecture: x86
- Attestors used: slsa, git, product, material, oci, gitlab
- Archivista version: not used

stupendoussuperpowers:

The current OCI attestor relies on `docker save` based `manifest.json` files to build up the attestation. This is not a file that is required or expected in an OCI image. 

As the issue points out, the current behaviour breaks for OCI compliant images (built and saved using for instance with `buildah --format oci`) error out since they do not have a `manifest.json`.

This PR makes the following changes:
- Detects `index.json` and uses that as the primary source for building the attestation.
- To preserve backwards compatibility with `docker save`, it falls back to `manifest.json` in case `index.json` is missing. 
- Adds a proper parser for OCI manifests, say for instance having the ability to traverse nester indices and detecting runnable images.
- Preserves the current schema for an OCI attestor (testing confirms that manually traversal of `index.json` does report the same things as reading the `manifest.json` for docker images)
- Adds unit-test for the new parser.

Notes:
The current schema for OCI relies heavily on what a docker `manifest.json` looks like, this is not super-ideal. For instance, when an image has multiple manifests, we do not do a great job of reporting those. We also do not have a good way of reporting non-runnable layers that the image might contain (e.g. `application/vnd.in-toto+json`, `application/spdx+json`)

We should look into updating the schema for an OCI attestor, and having a discussion about what the backward-compatibility options should be. (CCing @Vyom-Yadav for this)

Fixes #686 

## Acceptance Criteria Met

- [x] Docs changes if needed
- [X] [DCO Sign-off](https://github.com/apps/dco)

### DIFF
--- a/attestation/oci/oci.go
+++ b/attestation/oci/oci.go
@@ -24,12 +24,14 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/in-toto/go-witness/attestation"
 	"github.com/in-toto/go-witness/cryptoutil"
 	"github.com/in-toto/go-witness/log"
 	"github.com/invopop/jsonschema"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 const (
@@ -148,7 +150,7 @@ func (a *Attestor) Attest(ctx *attestation.AttestationContext) error {
 		return err
 	}
 
-	if err := a.parseMaifest(ctx); err != nil {
+	if err := a.parseManifest(ctx); err != nil {
 		log.Debugf("(attestation/oci) error parsing manifest: %w", err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


stupendoussuperpowers:
CC: @jkjell / @mikhailswift (Not sure who the best person to review this is)

mikhailswift:
i think this seems reasonable

mikhailswift:
as far as backward compatibility goes and changes to this attestor's schema -- i agree that the current shape of this attestor is not ideal. we tehcnically have the capability to version attestors and would be able to persist this as a v0.1 of the attestor -- but that capability still has gaps and hasn't been fully implemented. i'd pitch that this would be a good attestor to solve that with in a follow up

mikhailswift (approved):

PR #758, opened 2026-06-14
Fix ptrace and bpf logic for multithreaded processes

Vyom-Yadav:

**Migrates from `PTRACE_TRACEME` to `PTRACE_SEIZE` to support signal handling (to be used later by networktrace) using `PTRACE_LISTEN` - signal handling can't be done using `PTRACE_TRACEME`**

**https://stackoverflow.com/questions/63954630/how-to-handle-sigstop-sigcont-in-a-ptracer**

- Fixes handling of execve(2) for multi threaded processes which causes swap of thread group leader and kills existing other threads: 

> execve(2) under ptrace
>        When one thread in a multithreaded process calls [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html), the
>        kernel destroys all other threads in the process, and resets the
>        thread ID of the execing thread to the thread group ID (process
>        ID).  (Or, to put things another way, when a multithreaded process
>        does an [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html), at completion of the call, it appears as though
>        the [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html) occurred in the thread group leader, regardless of
>        which thread did the [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html).)  This resetting of the thread ID
>        looks very confusing to tracers:
> 

- Updated BPF logic to not rely on MAX_PARENT_WALK instead track TIDs of tasks as they are forked, cloned, etc. This also allows handling orphan processes that have been detached from the parent and are reparented under PID 1 where the earlier logic would've broken.
- BPF code now tracks the task lifecycle. 
- Added support for SIGSTOP and group-stops.
- Tests added. 

## Acceptance Criteria Met

- [ ] Docs changes if needed


attestation/commandrun/tracing_linux.go
```diff
@@ -117,8 +116,7 @@ func (p *ptraceContext) runTrace() error {
 		log.Infof("Entering hooks-only mode")
 		if p.hasPreExit {
 			log.Infof("Waiting for process exit to run PreExit hooks")
-			// Change the signal handling in waitForExitOnly if ptrace options are changed from PTRACE_O_TRACEEXIT
-			if err := unix.PtraceSetOptions(p.traceePid, unix.PTRACE_O_TRACEEXIT); err != nil {
+			if err := unix.PtraceSetOptions(p.traceePid, unix.PTRACE_O_TRACEEXIT|unix.PTRACE_O_TRACEEXEC|unix.PTRACE_O_TRACECLONE|unix.PTRACE_O_TRACEFORK|unix.PTRACE_O_TRACEVFORK); err != nil {
```
Vyom-Yadav: Just read, that all CLONE, FORK, VFORK start with a SIGSTOP. Need to handle that.

Vyom-Yadav: I have now fixed that.

jkjell (approved):

PR #766, opened 2026-06-24
BPF reproducibility GHA workflow

jkjell:

Add CI Checks for binary reproducibility of BPF binary files committed to the repo

This requires vmlinux.h to be committed to the repo to avoid kernel differences in CI runners (and locally).
Also updates `make generate-bpf` to use a docker command with the `cilium/ebp-builder` image.

## Acceptance Criteria Met

- [ ] Docs changes if needed
- [ ] Testing changes if needed


Vyom-Yadav:
I don't think vmlinux.h should be vendored. It should be generated in the workflow, used to compile to bpf bytecode and then deleted. For reproducibility, kernel version should be pinned instead of checking in vmlinux.h. If kernel version cannot be pinned, then reproducibility should include the kernel version and language toolchain versions. 

For checking whether they binary artifacts were not tampered with, we should try to do a binary diff using diffoscope or some custom reverse engineering logic.

PR #768, opened 2026-06-28
fix(environment): skip malformed glob patterns to avoid nil-glob panic

arpitjain099:

The environment attestor runs by default during `witness run`, and a malformed glob in a sensitive-key entry (say `*[`) crashes the whole attestation with a nil pointer dereference, so you lose all output.

Both ObfuscateEnvironmentArray and FilterEnvironmentArray compile a list of globs. When glob.Compile errors, the code logs it but still appends the nil compiled glob, and the later glob.Match(key) panics on it:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/in-toto/go-witness/environment.ObfuscateEnvironmentArray(...)
	environment/obfuscate.go:49
```

This is reachable from the public WithAdditionalKeys option and the `--attestor-environment-add-sensitive-key` CLI flag, so one bad user value brings down the run. I added a `continue` in the compile-error branch of both functions so an invalid pattern is logged and skipped and the nil glob is never stored. While there I fixed the filter.go log message, which still said "obfuscate glob pattern" (copy-paste from obfuscate.go).

## Verification

Added table cases to TestCapture_Capture for a malformed glob in both the obfuscate and filter paths; both panic on current code and pass here. `go test ./environment/...` passes, vet and gofmt clean.

## Acceptance Criteria Met

- [ ] Docs changes if needed

**Special notes for your reviewer**: Minimal change, one `continue` per function plus the log fix. Happy to tweak the wording or tests. Thanks!

### DIFF
--- a/environment/environment_test.go
+++ b/environment/environment_test.go
@@ -170,6 +170,49 @@ func TestCapture_Capture(t *testing.T) {
 				"TEST_TEXT": "value",
 			},
 		},
+		{
+			// A malformed glob (unterminated character class) used to panic
+			// because the failed compile result was still appended to the
+			// match list and later dereferenced. It should now be skipped.
+			name: "Obfuscate with malformed glob does not panic",
+			fields: fields{
+				sensitiveVarsList: DefaultSensitiveEnvList(),
+				addSensitiveVarsList: map[string]struct{}{
+					"*[": {},
+				},
+				excludeSensitiveVarsList:    map[string]struct{}{},
+				filterVarsEnabled:           false,
+				disableSensitiveVarsDefault: true,
+			},
+			args: args{
+				env: []string{
+					"TEST_TEXT=value",
+				},
+			},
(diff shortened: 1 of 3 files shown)


jkjell:
@matglas and others, I wonder if it's more appropriate to fail here as opposed to logging an error and continuing. This seems like the best way to recognize the intent to filter out or obfuscate a set of variables.

arpitjain099:
Good point @jkjell. For something whose whole job is filtering and obfuscating sensitive variables, failing on a malformed pattern is the safer default (fail closed), since silently skipping a bad glob could leave a variable you meant to hide exposed. My PR skips and continues just to stop the nil-glob panic, but I'm glad to change it to return an error on an invalid pattern instead, so a bad glob surfaces loudly rather than passing through. @matglas happy to go whichever way you both prefer, just let me know.

PR #772, opened 2026-06-30
fix(tracing): properly cleanup ENOENT files for child processes

### LINKED ISSUE #771: [Bug]: Tracer fails to clean up unresolvable ENOENT files for child processes
## Describe the bug
When using the `go-witness` tracer (e.g., via `witness run --trace`), files that do not exist (`ENOENT`) but are accessed by child processes are incorrectly recorded in the final attestation's `openedfiles` block as empty objects.

A common example of this is `systemd` NSS user/group lookups (like querying `/etc/userdb/<user>.user`). Since glibc automatically triggers these lookups for child processes (such as `tar` or `gpg`), the tracer records the paths. However, because the files don't actually exist on the host, they are never resolved to a hash. 

While the tracer does have a cleanup loop (`retryOpenedFiles`) meant to purge these unresolved paths, it currently only cleans up the `OpenedFiles` map belonging to the root tracee PID. As a result, the unresolved paths from child processes leak into the final attestation, making builds non-deterministic and unnecessarily exposing host specific `userdb` lookup paths.

## To Reproduce
1. Run a command with `--trace` that spawns a child process which attempts to read a nonexistent file. For example:
   ```bash
   witness run --trace -o test.json -- sh -c "cat /tmp/this_file_does_not_exist || true"
   ```
2. Decode the resulting attestation payload:
   ```bash
   jq -r '.payload' test.json | base64 -d > decoded.json
   ```
3. Inspect `decoded.json`. You will see `/tmp/this_file_does_not_exist` present as an empty object within the `openedfiles` block, despite it never existing on the filesystem.

## Expected behavior
The attestation should not include nonexistent files that were intercepted but could not be successfully read/hashed. The `retryOpenedFiles` function should properly remove all unresolvable entries for *all* tracked processes, not just the root tracee.

## Root Cause & Proposed Fix
The issue lies in `attestation/commandrun/tracing_linux.go` where `retryOpenedFiles()` only operates on `p.traceePid`:
```go
procInfo := p.getProcInfo(p.traceePid)
// loops over procInfo.OpenedFiles
```
The proposed fix is to update this logic to iterate over all tracked processes (`p.processes`) to ensure child process `OpenedFiles` maps are also cleaned up before finalization.

## Environment:
- **Go-Witness version:** v0.10.0 (or current main)
- **Architecture:** linux/amd64 (or specify applicable architectures)
- **Attestors used:** `commandrun` (with `--trace`)
- **Archivista version:** N/A (issue is within the generated attestation payload itself)

Jaydeep869:

**Description**

This PR fixes a bug in the `go-witness` tracer where unresolvable nonexistent files (`ENOENT`), such as `systemd` NSS user lookups (e.g., `/etc/userdb/lukp.user`), were leaking into the final attestation's `openedfiles` block as empty objects.

The tracer already has a `retryOpenedFiles()` cleanup loop designed to scrub these unresolved paths. However, it was previously only iterating over the root tracee's `OpenedFiles` (`p.traceePid`). This PR updates `retryOpenedFiles()` to iterate over all tracked processes (`p.processes`). This ensures that if a child process (like `gpg` or `tar`) triggers an unresolvable OS level file lookup, it is properly cleaned up from the final payload.

As part of this fix:
- The temporary `isUserdbPath` filter has been removed, since addressing the core cleanup loop makes explicit path filtering unnecessary.
- Added `Test_tracingChildProcessENOENT` regression test to ensure ENOENT paths accessed by child processes are consistently scrubbed.

## Which issue(s) this PR fixes

Fixes #771 

## Acceptance Criteria Met

- [ ] Docs changes if needed

### DIFF
--- a/attestation/commandrun/tracing_linux.go
+++ b/attestation/commandrun/tracing_linux.go
@@ -498,4 +498,3 @@ func getSpecBypassIsVulnFromStatus(status []byte) bool {
 
 	return false
 }
-
--- a/attestation/commandrun/tracing_linux.go
+++ b/attestation/commandrun/tracing_linux.go
@@ -289,21 +289,21 @@ func (p *ptraceContext) waitForExitOnly() error {
 
 func (p *ptraceContext) retryOpenedFiles() {
 	// after tracing, look through opened files to try to resolve any newly created files
-	procInfo := p.getProcInfo(p.traceePid)
+	for _, procInfo := range p.processes {
+		for file, digestSet := range procInfo.OpenedFiles {
+			if digestSet != nil {
+				continue
+			}
 
-	for file, digestSet := range procInfo.OpenedFiles {
-		if digestSet != nil {
-			continue
+			newDigest, err := cryptoutil.CalculateDigestSetFromFile(file, p.hash)
+
+			if err != nil {
+				delete(procInfo.OpenedFiles, file)
+				continue
+			}
+
+			procInfo.OpenedFiles[file] = newDigest
 		}
-
-		newDigest, err := cryptoutil.CalculateDigestSetFromFile(file, p.hash)
-
-		if err != nil {
-			delete(procInfo.OpenedFiles, file)
-			continue
-		}
-
-		procInfo.OpenedFiles[file] = newDigest
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jkjell (approved):
Thanks for this fix!

PR #809, opened 2026-08-03
[cryptoutil] `CalculateDigestSetFile` Fix blocking on FIFO, remove dead if-branches for hashablilty tests.

stupendoussuperpowers:

The current test for checking if a filepath is hashable has the following problems that this PR fixes:

1. It blocks indefinitely on FIFOs. This is because before it runs `isHashableFile` it calls `open(O_RDONLY)` on the file which does not immediately return for FIFOs. Since FIFOs are rejected anyway we use `O_NONBLOCK` so that open does not block.

2. `isHashableFile` has a lot of dead paths that are never executed: `mode.Perm().IsDir()` is always false since `Perm()` strips directory identifiers. `ModeSymLink` is not populated since the file is already opened. Checking just for IsRegular meets the same functionality as the current function.

(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*

Fixes #

## Acceptance Criteria Met

- [x] Docs changes if needed

### DIFF
--- a/cryptoutil/digestset_test.go
+++ b/cryptoutil/digestset_test.go
@@ -16,6 +16,9 @@ package cryptoutil
 
 import (
 	"crypto"
+	"os"
+	"path/filepath"
+	"syscall"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


Vyom-Yadav:
>        Normally, opening the FIFO blocks until the other end is
>        opened also.
> 
>        A process can open a FIFO in nonblocking mode.  In this case,
>        opening for read-only succeeds even if no one has opened on the
>        write side yet and opening for write-only fails with ENXIO (no
>        such device or address) unless the other end has already been
>        opened.
> 
>        Under Linux, opening a FIFO for read and write will succeed both
>        in blocking and nonblocking mode. 

This should be fine, it still opens for read-only mode with non blocking.

Vyom-Yadav (approved):
lgtm, but a test would make me happier

cryptoutil/digestset_test.go
```diff
@@ -76,3 +79,47 @@ func TestDigestSetEqual_EdgeCases(t *testing.T) {
 		})
 	}
 }
+
+func TestIsHashableFile(t *testing.T) {
+	regular := filepath.Join(t.TempDir(), "regular")
+	require.NoError(t, os.WriteFile(regular, []byte("hello"), 0o644))
+
+	fifo := filepath.Join(t.TempDir(), "fifo")
+	require.NoError(t, syscall.Mkfifo(fifo, 0o600))
```
mikhailswift, line 88: this will not work on windows, if this test is specific to linux it needs to be behind a build tag

PR #819, opened 2026-08-10
policy: predicate-type check in checkFunctionaries never gates acceptance

shmulc8:
### Summary

The predicate-type check in `Step.checkFunctionaries` (`policy/policy.go`) records a rejection but never gates acceptance, so a statement whose predicate type is not `attestation.CollectionType` can still land in `result.Passed`.

```go
if statement.Statement.PredicateType != attestation.CollectionType {
    result.Rejected = append(result.Rejected, RejectedCollection{...})
}   // no `continue` — falls through to the functionary check below
...
result.Passed = append(result.Passed, statements[i])   // reached even after the rejection above
```

Because `StepResult.Analyze()` passes when `len(Passed) > 0`, the check has no effect on the result — the same statement ends up in both `Rejected` and `Passed`.

### Is this a security issue?

I don't believe it's an exploitable bypass, and I don't want to overstate it. `validateAttestations` still requires the collection name to match the step and the expected attestation types to be present, and the envelope must be signed by an authorized functionary. A genuine non-collection predicate unmarshals to an empty `Collection` and is rejected there. So in practice this reads as dead defense-in-depth / a clarity issue rather than a vulnerability.

### Change

- Add `continue` so a non-collection predicate type is actually rejected.
- Add `TestCheckFunctionariesRejectsNonCollectionPredicate`, asserting such a statement never reaches `Passed` even with a valid functionary. The test fails on `main` (the statement lands in `Passed`) and passes with the one-line change.

If you'd instead prefer to keep relying on the `validateAttestations` backstop, this check is effectively dead code and I'm happy to replace it with a comment instead — whichever you'd like.

`go test ./policy/` passes.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -409,6 +409,7 @@ func (step Step) checkFunctionaries(statements []source.CollectionVerificationRe
 		// Check that the statement contains a predicate type that we accept
 		if statement.Statement.PredicateType != attestation.CollectionType {
 			result.Rejected = append(result.Rejected, RejectedCollection{Collection: statement, Reason: fmt.Errorf("predicate type %v is not a collection predicate type", statement.Statement.PredicateType)})
+			continue
 		}
 
 		if len(statement.Verifiers) > 0 {
--- a/policy/policy_test.go
+++ b/policy/policy_test.go
@@ -610,6 +610,45 @@ func TestCheckFunctionaries(t *testing.T) {
 	}
 }
 
+// A statement whose predicate type is not a witness collection must never be accepted,
+// even when it carries a valid signature from an authorized functionary. The predicate-type
+// check in checkFunctionaries must gate acceptance, not merely record a rejection reason.
+func TestCheckFunctionariesRejectsNonCollectionPredicate(t *testing.T) {
+	_, verifier, _, err := createTestKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	keyID, err := verifier.KeyID()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	step := Step{
+		Name: "step1",
+		Functionaries: []Functionary{
+			{Type: "PublicKey", PublicKeyID: keyID},
+		},
+		Attestations: []Attestation{
+			{Type: "dummy-prods"},
+		},
+	}
+
+	statements := []source.CollectionVerificationResult{
+		{
+			Verifiers: []cryptoutil.Verifier{verifier},
+			CollectionEnvelope: source.CollectionEnvelope{
+				Statement: intoto.Statement{PredicateType: "https://example.com/not-a-collection/v1"},
+			},
+		},
+	}
+
+	result := step.checkFunctionaries(statements, nil)
+
+	assert.Empty(t, result.Passed, "a non-collection predicate type must not be accepted even with a valid functionary")
+	assert.NotEmpty(t, result.Rejected, "a non-collection predicate type must be rejected")
+}
+
 // A single distinct collection that matches on every search-depth iteration must be counted once,
 // not once per iteration.
 func TestVerifyDeduplicatesCollectionsAcrossDepth(t *testing.T) {


0rkz:
Nice catch, and a clean fix — verified this against `0b3fa0d` (current head).

Confirmed the control-flow claim exactly: at `policy/policy.go:409-411` on `main`, the predicate-type check appends to `result.Rejected` with no `continue`/`return`, so execution falls into the functionary loop below and the same statement can still reach `result.Passed` at line 432 (`checkFunctionaries` doesn't return early). Your `continue` at line 412 closes it, and `TestCheckFunctionariesRejectsNonCollectionPredicate` reproduces the pre-fix failure and passes post-fix — I ran both.

I want to push back gently on one thing in the PR description, though, because I think it changes how this should be described. You wrote that a *genuine* non-collection predicate "unmarshals to an empty `Collection` and is rejected there" — that part I agree with, and I'm granting your other premise too, that the envelope has to be signed by an authorized functionary. The gap is that nothing in the path enforces genuineness: `envelopeToCollectionEnvelope` (`source/source.go:37`) unmarshals `Statement.Predicate` into `attestation.Collection{}` unconditionally — it never inspects `Statement.PredicateType` before doing so. So a statement whose declared predicate type is something else entirely, but whose predicate body is *shaped* like a collection (`{"name": ..., "attestations": [{"type": ..., "attestation": ...}]}`) naming the step and a registered attestation type, unmarshals into a fully populated, non-empty `Collection`. The collection-name and attestation-type checks in `validateAttestations` then match on attacker-chosen fields, so there's nothing left to catch it.

I checked this end-to-end rather than just at the `checkFunctionaries` level: built a real DSSE-signed statement (`dsse.Sign` + a real signer/verifier pair) with `PredicateType` set to an arbitrary non-collection URI and a forged-but-well-formed collection body, loaded it through `source.MemorySource.LoadEnvelope` → `source.VerifiedSource` (so the real, unexported `envelopeToCollectionEnvelope` parsing runs, not a hand-built struct), and called the public `Policy.Verify()` entrypoint against a step with no Rego policies (a common, realistic step shape).

On `main`: `pass=true`, `step1.Passed=1`, `step1.Rejected=1` — the same statement lands in both, and the top-level API reports the step as satisfied.
On your branch (`0b3fa0d`): `pass=false`, `step1.Passed=0`, `step1.Rejected=2` — the exploit no longer succeeds.

The step in that repro declares no `artifactsFrom`. I checked the other shape too, in case `verifyArtifacts` was doing the backstopping: when the step does declare one, the forged statement still rides along in `Passed` as long as one legitimate collection satisfies the edge, because `verifyArtifacts` only clears `Passed` when *nothing* satisfies it.

So this reads to me as a real bypass of the public verification API, not dead defense-in-depth — the one-line fix is doing real work, and I'd suggest dropping the "not exploitable" framing from the PR description (or at minimum not de-scoping it to a clarity issue) when it lands.

Separately: I grepped the whole module for the same shape (`.Rejected = append(...)` without a following `continue`/`return`/`break`) to check for recurrence elsewhere. Every other site is either inside a mutually-exclusive if/else with the `Passed` append (`policy.go:430`/`435`, `step.go:206`), immediately followed by `continue` (`policy.go:449`), or explicitly zeroes `Passed` in the same block (`policy.go:476`); the two remaining appends (`policy.go:260` and `:269`) only merge already-decided results and don't gate anything. None share this fall-through shape. So this fix is complete for the package as far as I can tell; no other spots need the same treatment.

*(Disclosure: I work on PayPerByte; the fail-open-check shape is one we've hit and fixed in our own verification path before, which is why I dug into the exploitability question. Drafted with AI assistance.)*

shmulik-apiiro:
You're right about `envelopeToCollectionEnvelope`, and that's a better argument for the `continue` than the one I made. My "unmarshals to an empty `Collection`" line only holds for a genuine foreign predicate, not for a forged collection-shaped body, so I'll drop the offer to replace the fix with a comment.

Before agreeing on impact I wanted the control case. Your repro signs with the step's authorized functionary key, so I added the same key signing a legitimate `CollectionType` statement with identical content, through the same path (`dsse.Sign` → `MemorySource.LoadEnvelope` → `VerifiedSource` → `Policy.Verify`):

```
                                                  main        #819
A  forged predicate type, authorized key      pass=true    pass=false
B  legit CollectionType, same key (control)   pass=true    pass=true
C  forged predicate type, unauthorized key    pass=false   pass=false
```

A reproduces your numbers exactly, including `Passed=1 Rejected=1` on main. B is the one I'd want on the record: an attacker who can satisfy the signature precondition already gets a passing verification on the patched branch, just by signing a real collection. The fall-through doesn't give them anything the key doesn't. C is there to show the functionary gate is what actually stops an outsider.

So I'd call this predicate-type confusion rather than a verification bypass. It bites where someone can get an authorized functionary to sign a statement whose predicate type they don't control but whose body they do, which is a real shape and enough for the check to earn its place. It isn't reachable unauthenticated, though, and I'd rather not have the PR claim that it is.

Thanks for the recurrence grep, that saved me a pass over the package. Happy to fold any of this into the description if the maintainers want it worded differently.

0rkz:
Agreed on both counts — predicate-type confusion is the right description, not verification bypass, and case B is the right control: an attacker who already holds an authorized key gets nothing here that the key doesn't already give them. Verified at `0b3fa0d`, not just read: case A is your own new test (`TestCheckFunctionariesRejectsNonCollectionPredicate`), rejects on the patched branch; the B-analogue is the existing `TestCheckFunctionaries` table's "simple 1 functionary pass" case, and it passes on both `main` and this branch; case C we constructed ourselves — the existing "invalid functionary" table case uses a legitimate `CollectionType` predicate, so it doesn't reach the predicate-type branch at all — and got Passed=0/Rejected=2 on `main`, Passed=0/Rejected=1 here, rejected either way. Strongest single result: lifting your case-A test onto `main` unmodified gives `Passed=1` *and* `Rejected=1` on the same statement at once; on `0b3fa0d` it's `Rejected=1` only. Full `policy` package suite green on both branches, no regressions from the fix.

(Disclosure: I work on PayPerByte. Drafted with AI assistance.)

PR #827, opened 2026-09-07
fix(gcp-iit): stop querying invalid project metadata routes

### LINKED ISSUE #258: Warning Issue with GCP Attestor
level=warning msg="failed to retrieve gcp metadata from http://metadata.google.internal/computeMetadata/v1/instance/project/project-id: unexpected status code for route: 404"
level=warning msg="failed to retrieve gcp metadata from http://metadata.google.internal/computeMetadata/v1/instance/project/numeric-project-id: unexpected status code for route: 404"

Eljees:
Fixes in-toto/witness#258.

`getInstanceData` queried the GCE metadata server for `project-id` / `project-number` at `InstanceMetadataUrl + "project/..."` — i.e. `/computeMetadata/v1/instance/project/project-id`, which is not a valid metadata route (project-level values live under `/computeMetadata/v1/project/`, with no `instance` segment). Every workload-identity attestation therefore logged a 404 warning for each of the two routes, which is exactly the noise reported in the issue.

Beyond the wrong URL, the fetched values were dead on arrival: a few lines later `getInstanceData` unconditionally overwrites `a.ProjectID` / `a.ProjectNumber` with the values parsed from the identity JWT by `parseJWTProjectInfo`, so even a successful fetch would have been discarded. The two routes are simply removed.

The endpoint map is extracted into a pure `buildInstanceEndpoints()` so the route set is testable without any HTTP: the regression test asserts the two project routes are gone and the six per-instance routes remain. It fails on the pre-fix map and passes here; `go test ./attestation/gcp-iit/`, `go vet` and `gofmt` are clean.


colek42:
Thank you!

colek42 (approved):

Eljees:
Thanks for the approval, @colek42. The two red checks here are not from this PR — both already fail on `main`, and this branch is based on exactly the commit they fail on (`9c7a11e`).

**`lint`** — staticcheck SA1019 in `signer/kms/azure/client.go:321-322`: `(crypto/ecdsa.PublicKey).X` and `.Y` were deprecated in Go 1.26. This PR's diff is two files under `attestation/gcp-iit/` and touches neither that package nor anything ECDSA.

**`Verify Schema`** — the committed schemas still carry `RawSignatureAlgorithm` on the embedded `x509.Certificate`, and current Go no longer has that field, so regenerating drops it. The whole diff in the job output is that one field disappearing from `jwt.json` and its neighbours; nothing in it comes from the gcp-iit attestor.

Same two jobs on `main@9c7a11e`, with no PR involved:

* lint — https://github.com/in-toto/go-witness/actions/runs/33822758254/job/100868795825
* Verify Schema — https://github.com/in-toto/go-witness/actions/runs/33822758169/job/100868794951

and on this branch:

* lint — https://github.com/in-toto/go-witness/actions/runs/34085158930/job/101800445823
* Verify Schema — https://github.com/in-toto/go-witness/actions/runs/34085158917/job/101800443003

Both look like the Go 1.26 toolchain bump rather than anything in the source: one deprecation that staticcheck now reports, one stdlib struct field that no longer exists when the schemas are regenerated. Happy to send a separate PR for either — regenerating the schemas is a one-command change, and the azure one is a small rewrite to `PublicKey.Bytes` — but they seemed out of scope to fold in here. Say the word and I will.

PR #836, opened 2026-09-16
Fix schema CI to use the Go version from go.mod

colek42:

`Verify Schema` fails on unchanged schemas because CI installs the Go version from `go.mod`, then ignores it by running generation in the rolling `cgr.dev/chainguard/go` image. Go 1.27 adds `crypto/x509.Certificate.RawSignatureAlgorithm`, producing diffs in the GCP IIT, GitHub, GitLab, and JWT schemas generated from the Go 1.26 toolchain.

Run verification directly with the toolchain already installed by `actions/setup-go`. Use the Go version from `go.mod` for the Linux image used by `make schema`, so local generation uses the same version and includes Linux-only attestors.

Observed failure: https://github.com/in-toto/go-witness/actions/runs/35115544284
Go library change: https://go.dev/doc/go1.27#cryptox509

The separate Azure `SA1019` lint failure is already fixed by #826, whose lint check passes. This PR complements that existing fix.

## Validation

- Linux schema verification with Go 1.26.4 (the version declared in `go.mod`).
- `actionlint .github/workflows/verify-schemagen.yaml`
- `make -n schema` selects `golang:1.26.4`.
- `git diff --check`

## Acceptance Criteria Met

- [x] DCO sign-off


stupendoussuperpowers:
`chainguard/go` was added as a part of #493 although it doesn't mention why this image in particular was chosen. 
@jkjell : Is that image load-bearing in any way? Currently it's breaking all schema GHAs.

colek42:
> `chainguard/go` was added as a part of #493 although it doesn't mention why this image in particular was chosen. @jkjell : Is that image load-bearing in any way? Currently it's breaking all schema GHAs.

This is just a quick fix.  We probably can build out own base image for this as the correct solution.

